Issue #9, opened 2018-11-29
Implement Baseline Integration Tests


sandysage:
Update the integration and accessibility tests to properly function for the "nsILogins" addon experience.

The [original tests](https://github.com/mozilla-lockbox/lockbox-extension/tree/master/test/integration) assume the functionality that requires a separate Firefox Accounts login and passwords database (from that built into Firefox).

Assume starting with a clean slate:
* add a login
* delete a login
* modify a login
* view the list on the management view
* view the list on the doorhanger

Parent of #54
Parent of #55
Parent of #56

PR #22, opened 2018-12-10
Refactor integration/accessibility tests

linuxwolf:
Connected to #9

This refactor better uses the Selenium-based `webdriver` and the underlying `web-ext` API to run integration functional/accessibility tests using the same patterns the unit tests use.

This also adds support for running accessibility testing via `axe-webdriverjs` (to start, the management homepage).

### DIFF
--- a/src/manifest.json.tpl
+++ b/src/manifest.json.tpl
@@ -8,6 +8,9 @@
     "96": "icons/lb_locked.svg"
   },
 
+
+  "content_security_policy": "script-src 'self' {{csp_scripts}} ; object-src 'self' {{csp_objects}}",
+
   "applications": {
     "gecko": {
       "id": "{{id}}",
--- a/src/manifest.json.tpl
+++ b/src/manifest.json.tpl
@@ -9,7 +9,7 @@
   },
 
 
-  "content_security_policy": "script-src 'self' {{csp_scripts}} ; object-src 'self' {{csp_objects}}",
+  "content_security_policy": "script-src 'self' {{testing_csp_scripts}} ; object-src 'self' {{testing_csp_objects}}",
 
   "applications": {
     "gecko": {
--- a/test/integration/a11y-test.js
+++ b/test/integration/a11y-test.js
@@ -1,0 +1,48 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import getWebExtension from "./driver";
+import AxeBuilder from "axe-webdriverjs";
+import chai from "chai";
+import chaiAsPromised from "chai-as-promised";
+
+chai.use(chaiAsPromised);
+
+describe("accessibility", () => {
+  let webext;
+  let helper;
+
+  before(async () => {
+    webext = await getWebExtension();
+    await webext.start();
+
+    helper = {
+      async home() {
+        const url = webext.url("/list/manage.html");
+        const { driver, webdriver } = webext;
+        await driver.get(url);
+
+        return driver.wait(webdriver.until.elementLocated(
+          webdriver.By.css("main#content")
+        ), 1000);
+      },
+    };
+  });
+  after(async () => {
+    await webext.stop();
+  });
+
+  beforeEach(async () => {
+    await webext.inContent();
+  });
+
+  it("tests homepage", async () => {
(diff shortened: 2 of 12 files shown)


6a68:
@linuxwolf On my mac, I'm not able to get `npm run integration` to correctly choose Nightly; it defaults to release channel Firefox, then the test fails, because the addon is unsigned. I think the binary needs to be explicitly set in the webdriver code (unless there's a command-line option I haven't discovered).

I found a reference snippet in a [semi-recent blog post](https://mykzilla.org/2017/08/30/headless-firefox-in-node-js-with-selenium-webdriver/) that seems relevant:

```js
const binary = new firefox.Binary(firefox.Channel.NIGHTLY);
binary.addArguments("--headless");
const driver = new Builder()
  .forBrowser('firefox')
  .setFirefoxOptions(new firefox.Options().setBinary(binary))
  .build();
```

This doesn't immediately work here, because the `firefox()` getter inside driver.js causes breakage when trying to access `firefox.Binary` and `firefox.Channel`. A little juggling should get it working, though, I think?

linuxwolf:
That's a good catch, and I'll add a way to set which Firefox to look for.

I lean toward making it configurable via environment variables and/or npm config properties, though.  For instance, I really want automated testing to run against either `unsigned` or `beta` so we're not too bleeding edge.

test/integration/driver.js
```diff
@@ -0,0 +1,142 @@
...
+
+    // load web-ext
+    const webext = this.webext;
+    await driver.setContext(firefox.Context.CHROME);
+    webext.id = await driver.installAddon(webext.extensionPath);
+    webext.hostname = await driver.executeScript(`
+      const Cu = Components.utils;
+      const Services = ChromeUtils.import("resource://gre/modules/Services.jsm");
+      const { WebExtensionPolicy } = Cu.getGlobalForObject(Services);
+      return WebExtensionPolicy.getByID("${webext.id}").mozExtensionHostname;
```
6a68: I'd probably ask somebody on the addons team (maybe @aswan?) if there's a better incantation to use here, just to be sure.

src/manifest.json.tpl
```diff
@@ -8,6 +8,9 @@
     "96": "icons/lb_locked.svg"
   },
 
+
+  "content_security_policy": "script-src 'self' {{csp_scripts}} ; object-src 'self' {{csp_objects}}",
```
6a68: Since this is a template file, not just straight json, maybe add a comment to explain that the CSP rule is needed to allow the a11y test to inject code into a running page

test/integration/a11y-test.js
```diff
@@ -0,0 +1,43 @@
...
+    await webext.inContent();
+  });
+
+  it("tests homepage", async () => {
+    await helper.management();
+
+    return AxeBuilder(webext.driver).
+      analyze().
+      then(results => {
+        expect(results).violations("critial").to.deep.equal([]);
```
6a68: typo: "critical"

linuxwolf: 🤦‍♂️

src/manifest.json.tpl
linuxwolf: While a template file, the template language isn't that sophisticated, but I'll see what I can do to approximate a comment.

test/integration/driver.js
linuxwolf: That's a good point.

This was copied almost verbatim from `lockbox-extension`'s test setup, and was a pattern I saw elsewhere, but it doesn't make it the best.

linuxwolf:
@mozilla-lockbox/desktop-engineering mind doing another CR?  There were some non-trivial changes to address the comments, and another pass seems worthwhile.

src/manifest.json.tpl
6a68: Feel free to ignore this suggestion if it makes things too complicated; future people can always rely on git blame to find their way to some context :-)

linuxwolf: it's kind of turning into that \-:

This project is using [json-templater](https://github.com/lightsofapollo/json-templater) to turn it into a full-fledged JSON file.  It supports a simplified mustache syntax via a hand-rolled parser.

I could replace this module with something else, but this was the best match we found the first and second times through.  The alternative seems to be to roll own own, which I'm really not inclined to do ...

linuxwolf: changes the `csp_` variable names to note they're for testing purposes

linuxwolf:
merged with @6a68 's earlier r+

PR #24, opened 2018-12-13
Bugfix for npm run run on Windows 10

lmorchard:
Use cross-env-shell to expand $npm_config_webext_runflags in `npm run run`


lmorchard:
Looks like cross-env isn't enough to get env vars expanded in npm script commands under Windows 10 powershell.

PR #41, opened 2018-12-18
Move clipboard operations into async background handler

### LINKED ISSUE #7: Clear clipboard after `copy`
## Acceptance criteria
* similar to mobile apps, should clear the clipboard 1 minute after a username and/or password is copied into it

## References
* [Clipboard API (MDN)](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API)

lmorchard:
- CopyToClipboardButton no longer does a clipboard operation, just
  displays "copied" feedback and passes resolved toCopy to onCopy

- Every onCopy handler now accepts toCopy to carry the clipboard value
  into the copiedField() action creator

- Tweak copiedField() action creator to dispatch COPIED_FIELD_STARTING and
  COPIED_FIELD_COMPLETED tracking async message to background script

- Add clipboard module to background script along with "copied_field"
  message type

- Add clipboardRead permission - so that we can check that clipboard
  content has not changed since we wrote to it before clearing it

- Drop copy-to-clipboard dependency

- Updates to tests

Fixes #7

### DIFF
--- a/src/background/clipboard.js
+++ b/src/background/clipboard.js
@@ -1,0 +1,24 @@
+const CLIPBOARD_CLEAR_DELAY = 1 * 60 * 1000; // 1 minute
+
+let clearClipboardTimer = null;
+
+export async function copyToClipboard(field, toCopy, navigatorClipboard = navigator.clipboard) {
+  if (clearClipboardTimer) {
+    // Clear any existing timer before scheduling a new one.
+    window.clearTimeout(clearClipboardTimer);
+  }
+  await navigatorClipboard.writeText(toCopy);
+  clearClipboardTimer = window.setTimeout(
+    async () => {
+      clearClipboardTimer = null;
+      // Only clear the clipboard if it still holds the value set earlier.
+      const inClipboard = await navigatorClipboard.readText();
+      if (inClipboard === toCopy) {
+        navigatorClipboard.writeText("");
+      }
+    },
+    CLIPBOARD_CLEAR_DELAY
+  );
+}
+
+export default { copyToClipboard };
--- a/package.json
+++ b/package.json
@@ -101,7 +101,6 @@
     }
   },
   "dependencies": {
-    "copy-to-clipboard": "^3.0.8",
     "fluent": "^0.8.1",
     "fluent-intl-polyfill": "^0.1.0",
     "fluent-langneg": "^0.1.0",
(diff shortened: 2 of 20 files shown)


src/background/clipboard.js
```diff
@@ -0,0 +1,21 @@
...
+
+export async function copyToClipboard(field, toCopy) {
+  if (clearClipboardTimer) {
+    // Clear any existing timer before scheduling a new one.
+    clearTimeout(clearClipboardTimer);
+  }
+  await navigator.clipboard.writeText(toCopy);
+  clearClipboardTimer = setTimeout(
+    async () => {
+      // Only clear the clipboard if it still holds the value set earlier.
```
lmorchard: This bit wasn't specified in the issue, but it's a thing that's bitten me in other password managers: If I copy a password, then go on to use the clipboard for other things later - the manager clears unrelated data from my clipboard unexpectedly. Had to add a clipboard read permission to get it working, which may or may not be a problem.

lmorchard:
Hmm, I guess that codecov failure means I should take a swing at writing some tests for the new background script clipboard module. Will do that

src/background/clipboard.js
linuxwolf: it wasn't called, it probably should have been called out, and it's great you called it out anyway (-:

linuxwolf:
> Hmm, I guess that codecov failure means I should take a swing at writing some tests for the new background script clipboard module.

What would be ideal; thanks!

src/background/clipboard.js
```diff
@@ -0,0 +1,23 @@
+const CLIPBOARD_CLEAR_DELAY = 1 * 60 * 1000; // 1 minute
+
+let clearClipboardTimer = null;
+
+export async function copyToClipboard(navigatorClipboard, field, toCopy) {
```
lmorchard: Wrestling a bit with getting a mock injected for the global `navigator.clipboard`. I just punted and kind of inject the dependency in params.

lmorchard:
Okay, I think I covered off on tests for the new clipboard module and the added message port handler

lmorchard:
Noticed I didn't have 100% coverage on the clipboard module, threw in one more test

src/background/clipboard.js
linuxwolf: this is a pattern we've followed if the "global" cannot be set or replaced.

I think I'd prefer for `navigatorClipboard` to be the last argument and default to `navigator.clipboard`.

lmorchard:
One more update to address the nit - can't think of why I didn't do it that way to begin with!

Issue #21, opened 2018-11-29
Use nsILoginInfo as the Data Model


linuxwolf:
Instead of using a custom data model, use the nsILoginInfo/MetaInfo model.  This considerably simplifies the implementation.

lmorchard:
Adding myself to assignees, since I think work toward this will be a natural direction along with the work to integrate the logins API (#4 / PR #47) that I'm about to try doing

lmorchard:
Seeing [my TODO comments](https://github.com/mozilla-lockbox/lockbox-addon/blob/10feb59a9194d400ccd49b4a9b548dd81b4ba972/src/background/datastore.js#L185) still in the code, I just realized that I left this issue only half done: The datastore is hooked up to the Logins API - but datastore is still doing internal conversions between the custom data model, use the nsILoginInfo/MetaInfo model. Going to reopen this and actually finish the job.

PR #47, opened 2019-01-04
Logins API

### LINKED ISSUE #4: Implement Logins WebExtension API (Browser.logins API)
A Logins API is necessary to function post Firefox 64 for the Lockbox add-on to access browser saved logins.

## Requirements
* Retrieve all login items
* Retrieve a specific item by ID
* Create a new login item
* Update an existing login item (by ID)
* Delete an existing login item by ID
* Notification when a login is changed (added, updated, removed)

## Risks
* Usability concerns
   * Prevent the extension’s web content (e.g., “management UX”) from triggering “save this login?” prompts (i.e., URL from Add extension’s URL to the list of disallowed hosts (for prompting to save logins)
* Security concerns
   * No permission needed to store extension’s “own” logins (by extension ID?)
   * Privileged addons + “logins”  permissions for access to all logins (PII and highly sensitive data!)
   * Including Favicons might be a privacy leak or concern?

6a68:
Fixes #4.

## Testing and Review Notes

This is just the logins API as an embedded API experiment--it doesn't yet connect to the rest of the addon code, and it's marked WIP because I haven't finished the tests yet.

Most of the documentation of the Login type in the schema comes straight from the nsILoginInfo and nsILoginMetaInfo interface definitions. Because LoginManager.update() allows any of the properties to be updated, I opted to add a second LoginUpdate type with all of the properties marked optional (and adding the convenient `timesUsedIncrement` shortcut). The webextension API schemas don't currently specify return types, so the return info is in the description.

I've opted to structure parts of the code in keeping with in-tree API conventions, though they seem kinda odd to me; I'll call those spots out inline.

-----

How to play around with the API:

1. Fire up nightly via `npm run run -- -f nightly`
2. Open up `about:debugging`
3. Click the 'Debug' button to open a debugger
4. In the debugger console, set up your listeners and a dummy Login by pasting in this code:

```js
login =  {
  guid: "{33535344-9cdb-8c4a-ae10-5849d0a2f04a}",
  timeCreated: 1546291981955,
  timeLastUsed: 1546291981955,
  timePasswordChanged: 1546291981955,
  timesUsed: 1,
  hostname: "https://example.com",
  httpRealm: null,
  formSubmitURL: "https://example.com",
  usernameField: "username",
  passwordField: "password",
  username: "creativeusername",
  password: "p455w0rd"
};
browser.experiments.logins.onAdded.addListener(login => console.log("onAdded fired: ", login))
browser.experiments.logins.onUpdated.addListener(login => console.log("onUpdated fired: ", login))
browser.experiments.logins.onRemoved.addListener(login => console.log("onRemoved fired: ", login))
```

5. Now you can play around with creating, updating, and removing logins:

```js
await browser.experiments.logins.getAll();
await browser.experiments.logins.add(login);
await browser.experiments.logins.get(login.guid);
await browser.experiments.logins.getAll();
login.username = "originalusername";
await browser.experiments.logins.update(login);
await browser.experiments.logins.touch(login.guid); // note: fires onUpdated
await browser.experiments.logins.remove(login.guid);
```

You can try to trigger errors by adding the login twice in a row, or removing it twice in a row, or setting incorrect field types when adding or updating, etc. You can also open up the login manager (`about:preferences#privacy`) and make changes there, then verify that the corresponding events fire in the debugger.

*Edited to change onCreated to onAdded in debugger setup script, and add touch() reference*

### DIFF
--- a/src/experiments/logins/api.js
+++ b/src/experiments/logins/api.js
@@ -49,7 +49,7 @@ this.logins = class extends ExtensionAPI {
           },
           add(loginInfo) {
             if (getLogin(loginInfo.guid)) {
-              throw new ExtensionError(`Add failed: Login already exists with GUID ${loginInfo.guid}`);
+              throw new ExtensionError(`Add failed: Login already exists with ID ${loginInfo.guid}`);
             }
             try {
               const login = LoginHelper.vanillaObjectToLogin(loginInfo);
@@ -63,7 +63,7 @@ this.logins = class extends ExtensionAPI {
           update(loginInfo) {
             const login = getLogin(loginInfo.guid);
             if (!login) {
-              throw new ExtensionError(`Update failed: Login not found with GUID ${loginInfo.guid}`);
+              throw new ExtensionError(`Update failed: Login not found with ID ${loginInfo.guid}`);
             }
             const loginAndMetaData = LoginHelper.newPropertyBag(loginInfo);
             Services.logins.modifyLogin(login, loginAndMetaData);
@@ -73,7 +73,7 @@ this.logins = class extends ExtensionAPI {
           remove(id) {
             const login = getLogin(id);
             if (!login) {
-              throw new ExtensionError(`Remove failed: Login not found with GUID ${id}`);
+              throw new ExtensionError(`Remove failed: Login not found with ID ${id}`);
             }
             try {
               Services.logins.removeLogin(LoginHelper.vanillaObjectToLogin(login));
--- a/src/experiments/logins/api.js
+++ b/src/experiments/logins/api.js
@@ -86,7 +86,7 @@ this.logins = class extends ExtensionAPI {
                 timeLastUsed: Date.now(),
               });
               Services.logins.modifyLogin(login, updates);
-              const updatedLogin = getLogin(loginInfo.guid);
+              const updatedLogin = getLogin(login.guid);
               return updatedLogin;
             } catch (ex) {
               throw new ExtensionError(ex);
--- a/src/experiments/logins/api.js
+++ b/src/experiments/logins/api.js
@@ -1,0 +1,164 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/* globals AppConstants, browser, Components, CustomizableUI, dispatcher,
+   ExtensionCommon, ExtensionAPI, Services, XPCOMUtils */
+
+"use strict";
+
+ChromeUtils.defineModuleGetter(this, "ExtensionUtils",
+                               "resource://gre/modules/ExtensionUtils.jsm");
+ChromeUtils.defineModuleGetter(this, "LoginHelper",
+                               "resource://gre/modules/LoginHelper.jsm");
+ChromeUtils.defineModuleGetter(this, "Services",
+                               "resource://gre/modules/Services.jsm");
+
+const { ExtensionError } = ExtensionUtils;
+
+const getLogins = () => {
+  const logins = Services.logins.
+    getAllLogins().
+    filter((l) => !(l.hostname || "").startsWith("chrome://")).
+    map(LoginHelper.loginToVanillaObject);
+  return logins;
+};
+
+const getLogin = (id) => {
+  const login = getLogins().
+    filter(l => l.guid === id)[0];
+  return login || null;
+};
+
+this.logins = class extends ExtensionAPI {
+  getAPI(context) {
+    const EventManager = ExtensionCommon.EventManager;
+    return {
+      experiments: {
+        logins: {
+          // See schema.json for function documentation.
+          getAll() {
+            const logins = getLogins();
+            return logins;
+          },
+          get(id) {
+            const login = getLogin(id);
+            return login;
+          },
+          // TODO: should this be called create(), or should the event name be changed to onAdded?
+          add(loginInfo) {
+            if (getLogin(loginInfo.guid)) {
+              throw new ExtensionError(`Add failed: Login already exists with GUID ${loginInfo.guid}`);
+            }
+            try {
+              const login = LoginHelper.vanillaObjectToLogin(loginInfo);
+              Services.logins.addLogin(login);
+            } catch (ex) {
+              throw new ExtensionError(ex);
+            }
+            const createdLogin = getLogin(loginInfo.guid);
+            return createdLogin;
+          },
+          update(loginInfo) {
+            const login = getLogin(loginInfo.guid);
+            if (!login) {
+              throw new ExtensionError(`Update failed: Login not found with GUID ${loginInfo.guid}`);
+            }
+            const loginAndMetaData = LoginHelper.newPropertyBag(loginInfo);
+            Services.logins.modifyLogin(login, loginAndMetaData);
+            const updatedLogin = getLogin(loginInfo.guid);
+            return updatedLogin;
+          },
+          remove(id) {
+            const login = getLogin(id);
+            if (!login) {
+              throw new ExtensionError(`Remove failed: Login not found with GUID ${id}`);
+            }
+            try {
+              Services.logins.removeLogin(LoginHelper.vanillaObjectToLogin(login));
+            } catch (ex) {
+              throw new ExtensionError(ex);
+            }
+          },
+          onCreated: new EventManager(context, "logins.onCreated", fire => {
+            const callback = (value) => {
+              fire.async(value);
+            };
+            const observer = {
+              observe: (subject, topic, type) => {
+                if (type !== "addLogin") {
+                  return;
+                }
+                subject.QueryInterface(Ci.nsILoginMetaInfo).QueryInterface(Ci.nsILoginInfo);
+                const login = LoginHelper.loginToVanillaObject(subject);
+                callback({ login });
+              },
+            };
+            Services.obs.addObserver(observer, "passwordmgr-storage-changed");
+            return () => {
+              Services.obs.removeObserver(observer, "passwordmgr-storage-changed");
+            };
+          }).api(),
+          onUpdated: new EventManager(context, "logins.onUpdated", fire => {
+            const callback = (value) => {
+              fire.async(value);
+            };
+            const observer = {
+              observe: (subject, topic, type) => {
+                if (type !== "modifyLogin") {
+                  return;
+                }
+                subject.QueryInterface(Ci.nsIArrayExtensions);
+                const newLogin = subject.GetElementAt(1);
+                const login = LoginHelper.loginToVanillaObject(newLogin);
+                callback({ login });
+              },
+            };
+            Services.obs.addObserver(observer, "passwordmgr-storage-changed");
+            return () => {
+              Services.obs.removeObserver(observer, "passwordmgr-storage-changed");
+            };
+          }).api(),
+          onRemoved: new EventManager(context, "logins.onRemoved", fire => {
+            const callback = (value) => {
+              fire.async(value);
+            };
+            const observer = {
+              observe: (subject, topic, type) => {
+                if (type !== "removeLogin") {
+                  return;
+                }
+                subject.QueryInterface(Ci.nsILoginMetaInfo).QueryInterface(Ci.nsILoginInfo);
+                const login = LoginHelper.loginToVanillaObject(subject);
+                callback({ login });
+              },
+            };
+            Services.obs.addObserver(observer, "passwordmgr-storage-changed");
+            return () => {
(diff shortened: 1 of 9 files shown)


src/experiments/logins/api.js
```diff
@@ -0,0 +1,164 @@
...
+ChromeUtils.defineModuleGetter(this, "ExtensionUtils",
+                               "resource://gre/modules/ExtensionUtils.jsm");
+ChromeUtils.defineModuleGetter(this, "LoginHelper",
+                               "resource://gre/modules/LoginHelper.jsm");
+ChromeUtils.defineModuleGetter(this, "Services",
+                               "resource://gre/modules/Services.jsm");
+
+const { ExtensionError } = ExtensionUtils;
+
+const getLogins = () => {
```
6a68: The existing APIs seem to move functions out to API-global scope, rather than have helper functions defined on the ExtensionAPI subclass, so I went with that approach for `getLogins` and `getLogin`.

src/experiments/logins/api.js
```diff
@@ -0,0 +1,164 @@
...
+              observe: (subject, topic, type) => {
+                if (type !== "addLogin") {
+                  return;
+                }
+                subject.QueryInterface(Ci.nsILoginMetaInfo).QueryInterface(Ci.nsILoginInfo);
+                const login = LoginHelper.loginToVanillaObject(subject);
+                callback({ login });
+              },
+            };
+            Services.obs.addObserver(observer, "passwordmgr-storage-changed");
```
6a68: Standard practice seems to be to wire up a separate observer for each event subscriber, and do the translation from Firefox-internal to API data type inside the callback, like I've done here. Seems odd to me, but I guess we really are only going to have one listener for each event (for now), so, ehh.

Related note: the old extension set up one listener, and then used a case statement to dispatch actions. With this logins API, the addon will wire up the redux store to the LoginManager via a bunch of event listeners:

```js
browser.experiments.logins.onCreated(login => {
  dispatcher.record({
    type: "login_changed",
    mode: "added",
    login: login,
  });
});
// same idea for onUpdated, onRemoved, onAllRemoved
```

src/experiments/logins/api.js
```diff
@@ -0,0 +1,164 @@
...
+            }
+            const createdLogin = getLogin(loginInfo.guid);
+            return createdLogin;
+          },
+          update(loginInfo) {
+            const login = getLogin(loginInfo.guid);
+            if (!login) {
+              throw new ExtensionError(`Update failed: Login not found with GUID ${loginInfo.guid}`);
+            }
+            const loginAndMetaData = LoginHelper.newPropertyBag(loginInfo);
```
6a68: Little bit subtle, but using a `nsIPropertyBag` has some [nice benefits](https://searchfox.org/mozilla-central/source/toolkit/components/passwordmgr/nsILoginManager.idl#77-86) and seems like the better choice for this API.

src/experiments/logins/api.js
```diff
@@ -0,0 +1,164 @@
...
+          },
+          remove(id) {
+            const login = getLogin(id);
+            if (!login) {
+              throw new ExtensionError(`Remove failed: Login not found with GUID ${id}`);
+            }
+            try {
+              Services.logins.removeLogin(LoginHelper.vanillaObjectToLogin(login));
+            } catch (ex) {
+              throw new ExtensionError(ex);
```
6a68: Error handling is still a little bit WIP here. If we don't rethrow errors as `ExtensionError`s, then the promise is still rejected, but the original error info is destroyed, and the webextensions code emits a particularly foul "An unexpected error occurred" message...but maybe we shouldn't expose all the errors in this way. Feedback welcome :-)

6a68:
BTW, [here](https://searchfox.org/mozilla-central/source/toolkit/components/extensions/parent/) is the directory in m-c that has the built-in APIs that run in the parent process. The schemas are [here](https://searchfox.org/mozilla-central/source/toolkit/components/extensions/schemas).

src/experiments/logins/schema.json
```diff
@@ -0,0 +1,260 @@
...
+      "pattern": "^\\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\}$"
+    }, {
+      "id": "Login",
+      "type": "object",
+      "description": "Represents a login, with the union of properties defined in either nsILoginInfo or nsILoginMetaData.",
+      "properties": {
+        "guid": {
+          "$ref": "Mozilla-GUID",
+          "optional": false,
+          "description": "The Mozilla-GUID to uniquely identify the login. This can be any arbitrary string, but a format as created by nsIUUIDGenerator is recommended. For example, '{d4e1a1f6-5ea0-40ee-bff5-da57982f21cf}'. (From nsILoginMetaInfo.idl)"
```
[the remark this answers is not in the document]
linuxwolf: I would relax the constraints now.  I don't think there's any old values to be worried about, but with future values.

The usefulness of GUID/UUID is starting to be questioned in the world, with the prevailing winds moving toward a hex-encoded string of a cryptographic hash process (starting with a cryptographically sufficient  source string).

src/experiments/logins/api.js
```diff
@@ -0,0 +1,164 @@
...
+    filter(l => l.guid === id)[0];
+  return login || null;
+};
+
+this.logins = class extends ExtensionAPI {
+  getAPI(context) {
+    const EventManager = ExtensionCommon.EventManager;
+    return {
+      experiments: {
+        logins: {
```
linuxwolf: I realize this is not in the current doc ...

We have found it helpful to have a `touch(id)` method that updates the `timeLastUsed` (and `timesUsed`).  It's not hard to duplicate that logic further in the stack, but it's a common enough use that I think it's valuable to add here if we could.

src/experiments/logins/api.js
lmorchard: With everything else declared in the schema, I'm almost surprised we don't have to specify the inventory of exceptions that could be thrown here rather than being able to clone-by-rethrow here. That's really the only thing I could think of to do differently - i.e. declare all the expected exceptions. But that feels possibly a little too boilerplatey

linuxwolf: It might require changes to the underlying nsILoginManager implementations to get better enumerations right now.

Eventually worth it, but for this initial work.

lmorchard:
Looks pretty good to me on a first read through the code and some console jiggery-pokery. I think I'll give a shot at actually hooking this up to Redux in the extension and see what I can do with it

linuxwolf:
>  think I'll give a shot at actually hooking this up to Redux in the extension and see what I can do with it

Currently Redux is in web content code, so this will likely need to go through the `background/message-ports` to keep it in the parent process.

src/experiments/logins/api.js
6a68: @lmorchard Yeah, the webextensions code in firefox doesn't validate or even parse the return values at all. I was told I could specify return types, but was afraid to do so: it's so easy to typo an unvalidated schema, and if they add return type checking, it might well break the API in the wild. I'm choosing to look at it as a "schema is half-full" situation ;-P

src/experiments/logins/api.js
```diff
@@ -0,0 +1,164 @@
...
+  const login = getLogins().
+    filter(l => l.guid === id)[0];
+  return login || null;
+};
+
+this.logins = class extends ExtensionAPI {
+  getAPI(context) {
+    const EventManager = ExtensionCommon.EventManager;
+    return {
+      experiments: {
```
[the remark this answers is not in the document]
6a68: > Is this namespace required? 

Yeah, the top-level 'experiments' namespace is [required](https://firefox-source-docs.mozilla.org/toolkit/components/extensions/webextensions/basics.html#webextensions-experiments) for all API experiments. 

> Will it cause problems with the other experimental API(s) we have planned?

Happily, no. I tried stubbing out a tiny `experiments.sync` API in the addon just now, and both were correctly merged and available in the debugger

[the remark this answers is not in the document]
linuxwolf: Awesome!  This would be worth adding to the [WebExt API dev docs](https://firefox-source-docs.mozilla.org/toolkit/components/extensions/webextensions/index.html) I think.

6a68:
Thanks for the great feedback! Changes pushed in tasty little commits. I'll re-request review and remove the WIP in the title when I've got the tests fully working and pushed to this branch.

6a68:
Rebased & force-pushed

linuxwolf:
>  I'll re-request review and remove the WIP in the title when I've got the tests fully working and pushed to this branch.

Works for me!  Also dismiss my previous review when you're ready.  I'll be sure to resolve the things that are done via discuss.

6a68:
@lmorchard OK, since you're poking at selenium things, here's tonight's puzzler: I can open the test page I've created, by doing the following:

- `NODE_ENV=test npm run run -- -f nightly` at the command line
- open up about:debugging
- in the console, do `browser.tabs.create({url: browser.extension.getURL("/test/integration/test-pages/logins-api.html")})`

And then, in the resulting test page, I can click the buttons and see stuff happen, and/or check for the existence of `browser.experiments.logins` in the console. So far, so good.

Now, if I run the new integration test†, there's a 30 second wait after that same test page loads, in which I can manually check that the buttons don't work, and looking in the console, I don't see anything under browser.experiments. What's up with that? No idea.

† There might be a cleaner incantation, but I've been running just this specific integration test via:
`./node_modules/.bin/cross-env NODE_ENV=test ./node_modules/.bin/mocha --require @babel/register --timeout 10000 test/integration/logins-api-test.js `

lmorchard:
> OK, since you're poking at selenium things

Well, I'm not actually doing that yet, and don't really have any advice there :/ Will probably dig into that after more work on #21 and getting this API working in the addon

6a68:
I'm going to try adding a test-only browser toolbar button. Maybe that'll open the page with the right permissions

lmorchard:
@6a68 Okay, so this is super underwhelming and I need to update tests, but I hooked up the Logins API to the datastore and it does seem to work in both directions for managing logins (i.e. from the add-on and from built-in options):

https://github.com/6a68/lockbox-addon/compare/logins-api...lmorchard:21-logins-api-hookup?expand=1

I should have had this working a day or two ago, but I kind of went down a premature rabbit hole chasing issue #21. 

I started playing with an overhaul replacing datastore with Redux store code shared between background script and frontend UI. Spent a little too long on that. Might still be a good idea, but going to shelve it for now and focus on getting these more minimal changes at least wrapped in unit tests.

test/integration/logins-api-test.js
```diff
@@ -0,0 +1,177 @@
...
+  });
+
+  it("browser.experiments.logins.update should update the username", async () => {
+    await addLogin();
+
+    await loadTestPage();
+    await clickButton("update");
+
+    const logins = await getLogins();
+    expect(logins[0].username).to.equal(mockLogin.username);
```
6a68: Oh, this is wrong. The username shouldn't match the initial value. I'll fix.

6a68: I've got something wrong with the `update()` method. Clearing review requests for now.

6a68:
Updated. I'm not sure why, but running `npm run integration`, I get a `SessionNotCreatedError` when starting the logins test. Maybe something in the functional test isn't getting cleaned up? Not sure, looking into it. For now, you can verify that logins test alone works by using the incantation

`./node_modules/.bin/cross-env NODE_ENV=test ./node_modules/.bin/mocha --require @babel/register --timeout 10000 test/integration/logins-api-test.js`

6a68:
Think I've got it. Should be green with this run

src/experiments/logins/api.js
```diff
@@ -0,0 +1,210 @@
...
+              throw new ExtensionError(ex);
+            }
+          },
+          update(newLogin) {
+            if (!("guid" in newLogin)) {
+              throw new ExtensionError(`Update failed: ID is required`);
+            }
+            // The addons framework code always supplies null values for missing keys,
+            // so filter those out.
+            // TODO: Could this cause subtle bugs? Would anyone ever want to update a field to be null? If so, maybe default to 'NULL" (the string), or undefined.
```
6a68: This is my one remaining point of uncertainty. Feedback welcome

linuxwolf: I did not realize this is the case.  This `null` behavior seems counter to "common idioms" with JSON-ish usage pattern (using `null` to clear a value, and `undefined` or omitted to ignore), arguably best embodied in [JSON merge patch](https://tools.ietf.org/html/rfc7386).

I would lean toward `newLogin` then having to be the complete representation rather than the subset that should actually be updated.

linuxwolf: To be  clear, I don't _think_ this behavior will have any impact on us in practice --- for now.  If we ever try to tackle toggling between "HTTP auth" and "submit form", this will cause problems.

6a68:
Thanks @linuxwolf! Filed #58 to address your final round of feedback. Squashing and merging.

Issue #49, opened 2019-01-07
Error in build command


meandavejustice:
`ERROR in Cannot read property 'readdir' of undefined` is the error I'm seeing after running
`npm install` and `npm run build`.

Looks like this error is coming from [here](https://github.com/mozilla-lockbox/lockbox-addon/blob/master/webpack.config.js#L54). Haven't figured out _which_ directory is undefined though.

```
Entrypoint list/popup = list/popup.js
[./background/browser-action.js] 804 bytes {background} [built]
[./background/index.js] 493 bytes {background} [built]
[./background/message-ports.js] 2.33 KiB {background} [built]
[./background/telemetry.js] 3.11 KiB {background} [built]
[./l10n.js] 2.94 KiB {list/manage} {list/popup} [built]
[./list/actions.js] 6.6 KiB {list/manage} {list/popup} [built]
[./list/manage/index.js] 1.13 KiB {list/manage} [built]
[./list/manage/reducers.js] 1.59 KiB {list/manage} [built]
[./list/manage/telemetry.js] 2.05 KiB {list/manage} [built]
[./list/message-ports.js] 878 bytes {list/manage} {list/popup} [built]
[./list/popup/containers/current-selection.js] 1.14 KiB {list/popup} [built]
[./list/popup/index.js] 1.51 KiB {list/popup} [built]
[./list/popup/reducers.js] 385 bytes {list/popup} [built]
[./list/popup/telemetry.js] 835 bytes {list/popup} [built]
[./telemetry.js] 361 bytes {list/manage} {list/popup} [built]
    + 192 hidden modules

ERROR in Cannot read property 'readdir' of undefined
Child html-webpack-plugin for "list/manage.html":
     1 asset
    Entrypoint undefined = list/manage.html
    [../node_modules/html-webpack-plugin/lib/loader.js!./template.ejs] 1010 bytes {0} [built]
        + 3 hidden modules
Child html-webpack-plugin for "list/popup.html":
     1 asset
    Entrypoint undefined = list/popup.html
    [../node_modules/html-webpack-plugin/lib/loader.js!./template.ejs] 1010 bytes {0} [built]
        + 3 hidden modules
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! lockbox@2.0.0-alpha build: `webpack`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the lockbox@2.0.0-alpha build script.
[...]

PR #50, opened 2019-01-08
Specify node version

meandavejustice:
- refs #49 

I was on an old node version that didn't support `fs.promises` which was making the build fail for me, adding this for other clumsy devs like myself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "version": "2.0.0-alpha",
   "description": "Take your passwords everywhere.",
   "engines": {
-    "firefox": ">= 63"
+    "firefox": ">= 63",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,7 +5,8 @@
   "version": "2.0.0-alpha",
   "description": "Take your passwords everywhere.",
   "engines": {
-    "firefox": ">= 63"
+    "firefox": ">= 63",
+    "node" : ">=11.6.0"
```
lmorchard: Is 11.6.0 the *minimum* version that works here? I've got 10.12.0 installed, and I think that's been working for me.

I'd kind of rather stick closer to the current LTS - which looks like 10.15.0 right now.

Issue #34, opened 2018-12-14
Edit Entry Details in Management Interface


linuxwolf:
I want to be able to easily change my entry details (e.g., username and/or password).

## Acceptance Criteria
* Ability to change username
* Ability to change password
* Can see the password value as it's edited
* Asked to discard changes if any field has changed but not yet saved
   - when clicking "Cancel"
   - when clicking another entry
* Asked to confirm delete

## Visual Design
**InVision**
https://mozilla.invisionapp.com/share/ENOL84959AK#/338187075_Logins_-_Entry_Detail_-_Edit_-Light-
Long passwords: https://mozilla.invisionapp.com/share/ENOL84959AK#/340867863_Logins_-_Entry_Detail_-_Edit_-_Long_Password_-Light-

**Zeplin**
https://zpl.io/VkwEO1M
Long passwords: https://zpl.io/adrvOB7

Issue #30, opened 2018-12-14
Error handling: Web Address must contain "https://" or "http://" 


sandysage:
The Web Address must contain a scheme (e.g., "https://") and a host (domain + port), and must not include a path.

If the URI is not well-formed, an error is displayed suggesting the user include "https://".

If the URI is well-formed but includes a path, the path can be stripped.

## Implementation Details

* Start using web URL class for validation.
* Extra testing to be sure URL class is sufficient.

## Visual Design
**InVision**
Helper: https://mozilla.invisionapp.com/share/ENOL84959AK#/339614514_Logins_-_New_Entry_-_Address_Focused_-Light-
Error: https://mozilla.invisionapp.com/share/ENOL84959AK#/339614519_Logins_-_New_Entry_-_Error_-_Address_-Light-

**Zeplin**
Helper: https://zpl.io/andLDWA
Error: https://zpl.io/aNKzXWk

linuxwolf:
We should only add "https://" if we're massaging data.  If we're informing the user of an error they need to correct, then we can indicate either, preferring "https://".

sandysage:
Convert to URI to strop off the end. Deliver error message if http:/ or https:/ is missing

Issue #18, opened 2018-11-29
Error Handling: "missing password" on add/edit save


sandysage:
The password cannot be empty.

## Visual Design
**InVision**
https://mozilla.invisionapp.com/share/ENOL84959AK#/339614518_Logins_-_New_Entry_-_Error_-_Password_-Light-

**Zeplin**
https://zpl.io/Vkweq3L

nickbrandt:
@sandysage is the open question just regarding the user choosing to Edit this entry, and then not being able to save it without a password? I think this is still fine to require the user to enter a password (or cancel out if they choose), as the entry won't be valuable without a password attached anyways.

sandysage:
For context, I deleted the open question.

PR #62, opened 2019-01-16
Initial minimal logins API connection to datastore

### LINKED ISSUE #21: Use nsILoginInfo as the Data Model

lmorchard:
Connected to #21

This doesn't actually fix #21, but it's kind of a start as it connects up the add-on to the real logins info data.

## Testing and Review Notes

* The add-on should now display real login entries
* Editing the login entries from the add-on should affect changes in the in-product login manager.
* Editing entries in the in-product login manager should affect changes in the add-on.

### DIFF
--- a/src/background/datastore.js
+++ b/src/background/datastore.js
@@ -4,9 +4,10 @@
 
 import UUID from "uuid";
 
+import { broadcast } from "./message-ports";
 import telemetry from "./telemetry";
 
-function convertInfo2Item(info) {
+export function convertInfo2Item(info) {
   let title;
   try {
     title = (new URL(info.hostname)).host;
@@ -36,7 +37,7 @@ function convertInfo2Item(info) {
   };
   return item;
 }
-function convertItem2Info(item) {
+export function convertItem2Info(item) {
   // dropped on the floor (for now ...)
   // * title
   // * tags
@@ -133,13 +134,16 @@ class DataStore {
       // assume formSubmitURL === hostname
       info.formSubmitURL = info.hostname;
     }
-    // TODO: call logins API
+
     let added = {
       ...info,
       guid: UUID(),
+      timesUsed: 0,
+      timeLastUsed: Date.now(),
       timeCreated: Date.now(),
       timePasswordChanged: Date.now(),
     };
+    await browser.experiments.logins.add(added);
     this._all[added.guid] = added;
 
     added = convertInfo2Item(added);
@@ -159,11 +163,11 @@ class DataStore {
       throw new Error(`no existing item for ${id}`);
     }
 
-    // TODO: call API
     let updated = {
       ...info,
       timePasswordChanged: Date.now(),
     };
+    await browser.experiments.logins.update(updated);
     this._all[updated.guid] = updated;
 
     updated = convertInfo2Item(updated);
@@ -174,7 +178,7 @@ class DataStore {
   async remove(id) {
     const item = await this.get(id);
     if (item) {
-      // TODO: call API
+      await browser.experiments.logins.remove(item.id);
       delete this._all[item.id];
 
       recordMetric("deleted", item.id);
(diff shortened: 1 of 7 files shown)


src/list/reducers.js
```diff
@@ -37,7 +37,13 @@ export function cacheReducer(state = {items: [], currentItem: null}, action) {
   case actions.ADD_ITEM_COMPLETED:
     return {
       ...state,
-      items: [...state.items, makeItemSummary(action.item)],
+      items: [
+        // HACK: add dispatched via Logins API events may be a duplicate
```
lmorchard: Small hacky wrinkle here: Since I'm broadcasting datastore changes based on Logins API events without any source, the UI source that made the changes will get its own change dispatched back to it. This will look like a duplicate. So, I papered over that by making the "add" action idempotent here.

lmorchard:
Okay, I *think* I worked out the issues. At least, insofar as I'm going to take off the "WIP" title and go get some lunch.

lmorchard:
Oh, but one other thing I may need to investigate for this PR or in a follow-up: Nothing here calls `touch()` on the Logins API to register that a login has been used. Wanted to do a little more digging to make sure I don't make fetches for display end up counting as a use.

src/background/datastore.js
```diff
@@ -133,13 +134,16 @@ class DataStore {
...
+
     let added = {
       ...info,
       guid: UUID(),
+      timesUsed: 0,
+      timeLastUsed: Date.now(),
       timeCreated: Date.now(),
       timePasswordChanged: Date.now(),
     };
+    await browser.experiments.logins.add(added);
```
6a68: This and other logins-modifying API methods do throw if something goes wrong, so I'd wrap them in try/catch blocks and just log the errors for now. I think we can handle the errors in the relevant issues (looks like #18, #30, #34 might all be relevant).

src/background/datastore.js
```diff
@@ -133,13 +134,16 @@ class DataStore {
...
     this._all[added.guid] = added;
```
6a68: I might be getting a little ahead of this transitional PR, but it seems cleaner to remove this line, so that `add()` just lobs the new login over to Firefox, and the datastore relies solely on the `onAdded` listener to update the list of logins.

src/background/datastore.js
```diff
@@ -159,11 +163,11 @@ class DataStore {
       throw new Error(`no existing item for ${id}`);
     }
 
-    // TODO: call API
     let updated = {
       ...info,
       timePasswordChanged: Date.now(),
     };
+    await browser.experiments.logins.update(updated);
     this._all[updated.guid] = updated;
```
6a68: Same comments as `add()` on these two lines (try/catch and maybe not directly updating the internal datastore).

src/background/datastore.js
```diff
@@ -174,20 +178,64 @@ class DataStore {
   async remove(id) {
     const item = await this.get(id);
     if (item) {
-      // TODO: call API
+      await browser.experiments.logins.remove(item.id);
       delete this._all[item.id];
```
6a68: Same comments as above for this spot, too.

test/unit/background/datastore-test.js
```diff
@@ -0,0 +1,266 @@
...
+
+    const apiRemove = browser.experiments.logins.remove;
+    expect(apiRemove.callCount).to.equal(1);
+    expect(apiRemove.lastCall.lastArg).to.equal(id);
+  });
+});
+
+const cmpAlphaBy = name => (a, b) => a[name].localeCompare(b[name]);
+
+const SAMPLE_INFOS = {
```
6a68: For completeness' sake, probably worth including the rest of the [`nsILoginMetaInfo`](https://searchfox.org/mozilla-central/source/toolkit/components/passwordmgr/nsILoginMetaInfo.idl) fields in these mocks: `timeCreated`, `timeLastUsed`, `timePasswordChanged`, `timesUsed`.

src/background/datastore.js
lmorchard: Hmm, I can just silently log them, though datastore.js *does* throw errors for other validation problems and such. Kind of figured it was useful to let these throw too, though it doesn't look like we do anything useful with them yet. (e.g. report an error over a message port, etc)

src/background/datastore.js
lmorchard: Yeah, that kind of feels like in the spirit of uni-directional dataflow ala redux, even if we're not using it here :)

linuxwolf:
> Nothing here calls `touch()` on the Logins API to register that a login has been used

That was on the list of things to get to from before; I'll make sure it's accounted for in our issues.

src/background/datastore.js
linuxwolf: Flux is a way of life (-:

src/background/message-ports.js
```diff
@@ -44,19 +44,16 @@ export default function initializeMessagePorts() {
     case "add_item":
       return openDataStore().then(async (ds) => {
         const item = await ds.add(message.item);
-        broadcast({type: "added_item", item}, sender);
```
lmorchard: Realized that since I've got the datastore now broadcasting these on the Logins API event handlers, all these messages got doubled up.

test/unit/background/message-ports-test.js
```diff
@@ -80,7 +83,6 @@ describe("background > message ports", () => {
     itemId = result.item.id;
 
     expect(result.item).to.deep.include(item);
-    expect(selfListener).to.have.callCount(0);
```
lmorchard: Since the datastore broadcasts messages to relay events from the Logins API, there's no sender or self available at that time to refrain from sending the event back to "self"

test/unit/mocks/browser.js
```diff
@@ -190,4 +194,22 @@ window.browser = {
     registerScalars(category, scalars) {},
     scalarSet(name, value) {},
   },
+
+  experiments: {
+    logins: {
+      async getAll() { return []; },
+      async add(login) {
+        browser.experiments.logins.onAdded.getListener()({ login });
```
lmorchard: I was hoping I could be less verbose here by just using `this`, but that apparently ends up being the wrong reference

lmorchard:
Addressed comments except for wrapping things in try/catch to log errors. We might want to let those throw, so we can catch those higher up along with datastore validation errors.

src/background/datastore.js
6a68: > datastore.js does throw errors for other validation problems and such

Ah, ok. Probably fine as is, then.

6a68:
Nice work :+1: Took this branch for a test drive, and all the logins add / update / delete happy paths are seamlessly mirrored across the logins management UI in Firefox, and the Lockbox management UI.

I dunno if @linuxwolf wants to take a pass before merging, but LGTM

test/unit/mocks/browser.js
linuxwolf: 🤔

PR #63, opened 2019-01-18
Implement Sync API (#23)

### LINKED ISSUE #23: Implement Sync WebExtension API (Browser.sync API)
A Sync API to function post Firefox 64 for the Lockbox add-on to access state information related to Firefox Sync.

## Requirements

* Determine if browser-wide sync is available (i.e., signed in to FxA)
* Determine if syncing of <type> is enabled (i.e., relevant pref is `true`).
* Provide sync profile information (i.e., FxA profile), if available
* Notification when profile information/state changes (in-browser changes)

## Risks

* security concerns
  * Privileged addons + “sync” permission for profile information (contains PII)
  * Sync state (inactive / disabled / enabled) privileged?
* implementation concerns
  * at least some portions of this are already defined in [`chrome.identity`](https://developer.chrome.com/extensions/identity); marked as "unsupported" within Firefox as of 65

6a68:
Fixes #23.

## Testing and Review Notes

I've omitted the rethrowing of Errors as `ExceptionError`s in this API, because the underlying code is a lot less likely to throw than LoginManager code was. I can add that boilerplate for consistency with the logins API; comments welcome.

I might have overcomplicated the integration test by extracting the DOM-twiddling code into helper methods. Comments welcome on that point (see the TODO in the file).

For some reason, I'm seeing a "Services is not defined" error when the integration test runs. I've sprinkled extra Services.jsm imports throughout the test, but still haven't tracked down the error. Cleaning that up is a harmless-yet-annoying TODO as well.

Finally, I still need to update the desktop requirements doc to match the changes discussed earlier this week. One further change I've made is to add a `SyncStatus` type, currently either "error" or "ok"; we could change this to expose the two sync errors ("login_failed", stale password error, and "not_verified", unverified email error), or it could just be a boolean and not a separate type. Feedback welcome, and really depends on how this will evolve in the future.

### How to play around with the API:

1. Fire up nightly via `npm run run -- -f nightly`
2. Open up `about:debugging`
3. Click the 'Debug' button to open a debugger
4. In the debugger console, set up your event listener by pasting in this code:

```js
browser.experiments.sync.onUserProfileChanged.addListener(uiState => {
  console.log("onUserProfileChanged fired: ", uiState)
});
```

In the debugger console, you can verify the current Sync/FxA state:

```js
await browser.experiments.sync.getUserProfileInfo();
// should initially return null
```

And you can also verify the state of the pref that controls whether Sync handles the logins saved in the browser (this can be changed by modifying the `sync.engines.passwords` pref directly, or by checking the checkbox in the about:preferences#sync screen):

```js
await browser.experiments.sync.getCurrentState("passwords");
// should initially return true
```

Now you can log in to the browser (hamburger menu > Sign In To Sync), and you'll both see updates from your event listener, as well as be able to query the `getUserProfileInfo` API method to see the state changes.

Note that a first-time login with an unverified email produces an error state. You'll also see a transition to an error state if you log in, then change the FxA password on the FxA website, causing the Firefox login to become stale. The API doesn't distinguish between these two error states right now, because in either case, we'll just send the user to `about:preferences#sync` to fix whatever's wrong.

Error states look like this:

```js
Object {
  id:  "fcea867fcea867fcea867fcea867fcea", 
  email: "jhirsch+asdf@mozilla.com", 
  displayName: null, 
  avatar: null, 
  status: "error" }
```

Verified logins look like this:

```js
{ status: "ok", 
avatar: "https://firefoxusercontent.com/146a19121ec0f71ed250a6a2fc0ff333", 
displayName: "🚨 Cheez Whiz 🚨",
email: "jhirsch+foo@mozilla.com",​
id: "..." }
```

### DIFF
--- a/src/experiments/sync/api.js
+++ b/src/experiments/sync/api.js
@@ -1,0 +1,99 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/* globals ChromeUtils, ExtensionAPI, ExtensionCommon, ObjectUtils, Services,
+   UIState */
+
+"use strict";
+
+ChromeUtils.defineModuleGetter(this, "Services",
+                               "resource://gre/modules/Services.jsm");
+ChromeUtils.defineModuleGetter(this, "UIState",
+                               "resource://services-sync/UIState.jsm");
+ChromeUtils.defineModuleGetter(this, "ObjectUtils",
+                               "resource://gre/modules/ObjectUtils.jsm");
+
+const getProfileInfo = async () => {
+  const uiState = UIState.get();
+  // STATUS_LOGIN_FAILED means the password was changed on another device, and the
+  // user needs to log in again.
+  // STATUS_NOT_VERIFIED means the user has logged in with an unverified email.
+  // In both cases, the user needs to take action elsewhere in Firefox.
+  const errors = [UIState.STATUS_LOGIN_FAILED, UIState.STATUS_NOT_VERIFIED];
+
+  let profileInfo;
+
+  // STATUS_NOT_CONFIGURED means the user is not logged in.
+  if (uiState.status === UIState.STATUS_NOT_CONFIGURED) {
+    profileInfo = null;
+  } else { // UIState.STATUS_SIGNED_IN, the user is logged in and verified.
+    const fxa = await UIState._internal.fxAccounts.getSignedInUser();
+    const isErrorStatus = errors.includes(uiState.status);
+
+    profileInfo = {
+      id: fxa && fxa.uid,
+      email: uiState.email,
+      displayName: uiState.displayName || null,
+      avatar: uiState.avatarURL || null,
+      status: isErrorStatus ? "error" : "ok",
+    };
+  }
+
+  return profileInfo;
+};
+
+this.sync = class extends ExtensionAPI {
+  getAPI(context) {
+    const EventManager = ExtensionCommon.EventManager;
+    return {
+      experiments: {
+        sync: {
+          async getCurrentState(syncType) {
+            return Services.prefs.getBoolPref("services.sync.engine." + syncType, false);
+          },
+          async getUserProfileInfo() {
+            const profileInfo = await getProfileInfo();
+            return profileInfo;
+          },
+          onUserProfileChanged: new EventManager(context, "sync.onUserProfileChanged", async (fire) => {
+            let oldValue = await getProfileInfo();
+            const callback = (value) => {
+              fire.async(value);
+            };
+            const observer = {
+              observe: async (subject, topic, type) => {
+                const newProfileInfo = await getProfileInfo();
+                // Sync fires duplicate events. Filter out events where:
+                // - old and new UserProfileInfo are both null
+                // - old and new UserProfileInfo are objects with same values.
+                // Happily, ObjectUtils.deepEqual(null, null) is true.
+                if (ObjectUtils.deepEqual(oldValue, newProfileInfo)) {
+                  return;
+                }
+                // Sync also fires two events when a verified login occurs:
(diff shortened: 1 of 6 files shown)


test/integration/sync-api-test.js
```diff
@@ -0,0 +1,307 @@
...
+
+      await UIState.refresh();
+    })().then(arguments[0]);
+    `);
+  };
+
+  const MockUserProfileInfo = {
+    loggedIn: {
+      avatar: "https://firefoxusercontent.com/00000000000000000000000000000000",
+      id: null, // TODO: why is the id null? am I setting this up wrong?
```
6a68: Think I've got a fix for this: need to borrow some machinery from an FxA-specific test, https://searchfox.org/mozilla-central/source/services/fxaccounts/tests/xpcshell/test_accounts_device_registration.js

6a68: This turns out to be [a lot of code](https://searchfox.org/mozilla-central/source/services/fxaccounts/tests/xpcshell/test_accounts_device_registration.js#94-128) to mock out, and I'm not quite sure it'll work. I'm inclined to leave FxA uid verification testing to a followup ticket. Thoughts?

6a68: eoger in #sync on Slack suggested I just use some form of dependency injection, as in https://searchfox.org/mozilla-central/rev/dac799c9f4e9f5f05c1071cba94f2522aa31f7eb/services/fxaccounts/FxAccountsProfileClient.jsm#35. I guess I could expose an fxa injector method in the API, like `testSetFxAccounts()`, since I have no idea how the WebExtensions framework code starts up the API itself...but that seems gross.

6a68:
I can't reproduce the test failures locally. The error is quite odd: `ChromeUtils is not defined`.

test/integration/sync-api-test.js
6a68: The problem, it turns out, was that I wasn't using the FxAccounts instance available on UIState._internal. I was loading FxAccounts.jsm.

6a68:
This is now fully ready for review. Hopefully we'll get a green test run from CI.

src/experiments/sync/api.js
```diff
@@ -0,0 +1,99 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/* globals AppConstants, browser, Components, CustomizableUI, dispatcher,
+   ExtensionCommon, ExtensionAPI, Services, XPCOMUtils */
+
+"use strict";
```
linuxwolf: Just a comment, nothing to take action on; I die a little on the inside every time I see it

src/experiments/sync/api.js
```diff
@@ -0,0 +1,99 @@
...
+                // - old and new UserProfileInfo are objects with same values.
+                // Happily, ObjectUtils.deepEqual(null, null) is true.
+                if (ObjectUtils.deepEqual(oldValue, newProfileInfo)) {
+                  return;
+                }
+                // Sync also fires two events when a verified login occurs:
+                // one as soon as login succeeds, with status "ok" but null
+                // avatar and displayName, and a second after pinging the FxA
+                // profile server, with status 'ok' and non-null avatar and
+                // possibly non-null displayName. Suppress the first one.
```
linuxwolf: 👍

src/experiments/sync/api.js
6a68: 😂 just cargo culting the [existing style](https://searchfox.org/mozilla-central/search?q=strict&case=false&regexp=false&path=toolkit%2Fcomponents%2Fextensions%2Fparent), and now that I look, I think I copypasta'd the eslint globals incorrectly anyway. I can get rid of this

6a68: Ah, on closer examination, there is a logical reason to use that directive: Mozilla code doesn't use ES6 modules, so strict mode isn't applied by default :-\

PR #64, opened 2019-01-22
Doorhanger list view updates

### LINKED ISSUE #12: Update doorhanger list view
When I want to take inventory of the accounts I have saved for a specific URL, I want an easy way to navigate the list to make decisions around next actions (view details, copy, manage).

## Acceptance criteria
* Should align to photon design standards
* Should feel native to the Firefox browser

## Visual Design
**Zeplin**
Default (search focused): https://zpl.io/2EPeOpy
Filtered: https://zpl.io/25G8Jln
Entry Hover: https://zpl.io/andg3Rk
Info Button Hover: https://zpl.io/bJ9JDAJ
Bottom Button ("Open Lockbox") Hover: https://zpl.io/V4QNvg4

meandavejustice:
Fixes #12

## Screenshots or Videos
list-button-hover
info-hover
default-list
filtered
list-hover

## To Do
- [X] Add banner with filtered item count in search view (waiting till sync work is done so I have access to data in component)


6a68:
@meandavejustice looks like this needs a rebase and a few linter fixes

meandavejustice:
info icon downloaded from zeplin looks bad for some reason
firefox_giewoz0af8

meandavejustice:
@nickbrandt @6a68 I believe all of your comments have now been addressed, I updated the pr description with screenshots also

nickbrandt:
@meandavejustice this is looking really close, thanks for addressing those items. Just noticing a few remaining things, identified below:

#### Entry List - Default
default

- Doorhanger background: #F2F2F4 (custom)
- Placeholder text: Gray 50
- Info icons/buttons: 16px padding from right side
- Info button: 32px width/height
- Header/Search height: 50px

---

#### Entry List - Filtered/Searching
filtered

- Text should be Grey 90
- When only 1 entry in results, text needs to read “1 entry found” (vs “entries”)
- Results text area should be similar to default state (background, font weight/color)
- Clear button: 16px padding from right side

---

#### Entry List - Hover States
hover

- Can we darken info icon to Gray 50 on hover of entry list item

---

#### Entry List - Hover States
screen shot 2019-01-29 at 11 24 03 am

Because the entry list items are tied to what you see in the full view, we are now seeing the info icons/buttons there as well. We actually don't need them there, can we update this for those to only be visible in the doorhanger?

Thanks!

nickbrandt:
@meandavejustice I'll try and look into the Info icon, it appears pretty close on my Mac, but does look a bit pixelated on your screenshot

src/list/components/item-summary.js
```diff
@@ -74,6 +74,7 @@ export default function ItemSummary({className, id, title, username, verbose,
                    $length={trimmedUsername.length}>
           <div data-name="subtitle" className={styles.subtitle}>no uSERNAMe</div>
         </Localized>
+        <div className={styles.info}></div>
```
linuxwolf: This inadvertently causes a breakage in `test/unit/list/components/item-summary-test.js` because more than one element matches `div > div`.

I think the most expedient way to fix htis is to change the [`merge classNames` test](https://github.com/mozilla-lockbox/lockbox-addon/blob/c36df2c25d65161c72c3eb5021a153879040d876/test/unit/list/components/item-summary-test.js#L47) to explicitly set an item id and then use it as `data-item-id` when `find()`ing the item.

6a68:
@meandavejustice Hey, when this is ready, remove the 'WIP' from the title, and re-request a review. Would be cool to fix the "merge master into branch" git noise, too.

src/list/popup/components/item-list-panel.js
```diff
@@ -67,35 +67,46 @@ export default function ItemListPanel({inputRef, noResultsBanner, ...props}) {
...
+    } else if (isFiltering) {
+      const count = props.items.length;
+      banner = (
+          <Localized id="filtered-banner" $count={count}>
+            <PanelBanner border="floating"> {count} eNTRIEs fOUNd</PanelBanner>
+          </Localized>
+      );
     } else {
-      topBorder = "floating";
+      banner = (
```
linuxwolf: this causes a breakage in `test/unit/list/popup/components/item-list-panel-test.js`.

There is now always a [`PanelBanner`](https://github.com/mozilla-lockbox/lockbox-addon/blob/c36df2c25d65161c72c3eb5021a153879040d876/test/unit/list/popup/components/item-list-panel-test.js#L90), where previously there were some cases it would not be there.

src/list/popup/components/item-list-panel.js
```diff
@@ -35,16 +35,16 @@ class PopupItemList extends React.Component {
   render() {
     const {items, ...props} = this.props;
     const {selected} = this.state;
-    const verbose = items.length <= MAX_VERBOSE_ITEMS;
 
     return (
-      <ItemList {...props} items={items} verbose={verbose} selected={selected}
+        <ItemList {...props} panel={true} items={items} verbose={false} selected={selected}
```
linuxwolf: This causes breaking in `test/unit/list/popup/containers/all-items-panel-test.js`.  We'd have to look back over the designs, but I think it's safe enough to remove `verbose` altogether.

Otherwise, [this check](https://github.com/mozilla-lockbox/lockbox-addon/blob/c36df2c25d65161c72c3eb5021a153879040d876/test/unit/list/popup/containers/all-items-panel-test.js#L100) should be removed.

nickbrandt:
@meandavejustice I couldn't find anything wrong with the info icon we were exporting through Sketch, so I went ahead and downloaded it from the [Photon site](https://design.firefox.com/icons/viewer/), and updated it to be the same color/sizing we were using (attached below). Let me know if this works out any better. It may be just some issue with exporting through Sketch and how that is getting translated (though that would be strange as other icons are working this way), but if you still have issues perhaps just try using the one straight from the Photon site and through CSS update the color/size. Let me know how it turns out!

[icon-info.svg.zip](https://github.com/mozilla-lockbox/lockbox-addon/files/2818885/icon-info.svg.zip)

meandavejustice:
Thanks @nickbrandt just pushed an update, your version of the info icon took care of it on my end.

nickbrandt:
@meandavejustice the caret looks good 👍 

Still seeing #737373 for the Search text. Looks like perhaps the empty state text got updated instead?

screen shot 2019-02-04 at 9 30 48 am
screen shot 2019-02-04 at 9 31 04 am
screen shot 2019-02-04 at 9 31 34 am

meandavejustice:
@nickbrandt God sorry, it should be actually correct now!

nickbrandt:
@meandavejustice no worries, looks good now. Thanks!

screen shot 2019-02-04 at 10 12 03 am

PR #66, opened 2019-01-24
 Integration tests for basic item add and view

### LINKED ISSUE #54: Implement View Integration Tests
Integration test of:

* Listing entries in the management interface

lmorchard:
Finding myself starting to spill over into writing other tests, so I figured I'd get this PR submitted and have eyes on it. Let's see what we think of the general approach here before I get much further.

Fixes #54 

Also includes code to hook up the `onAllRemoved` event that I'd missed in my previous PR for the Logins API.

## Testing and Review Notes
* Make sure you have [geckodriver](https://github.com/mozilla/geckodriver/releases) in your path
* `npm run integration`

### DIFF
--- a/test/integration/functional-test.js
+++ b/test/integration/functional-test.js
@@ -13,32 +13,261 @@ chai.use(chaiAsPromised);
 
 const ident = "lockbox_mozilla_com";
 
-describe("Lockbox functional testing", () => {
-  let webext, helper;
+// Note: Update these selectors if the DOM changes in components
+const makeSelectors = ({ By }) => ({
+  panelButton: By.id(`PanelUI-webext-${ident}-browser-action-view`),
+  addItemButton: By.id("addItemButton"),
+  newItemForm: By.id("newItemForm"),
+  newItemFormField: name => By.css(`#newItemForm input[name=${name}]`),
+  newItemFormSubmit: By.css("#newItemForm button[type=submit]"),
+  itemDetails: name => By.css(`span[data-name=${name}]`),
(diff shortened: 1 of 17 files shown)


test/integration/functional-test.js
```diff
@@ -13,32 +13,261 @@ chai.use(chaiAsPromised);
 
 const ident = "lockbox_mozilla_com";
 
-describe("Lockbox functional testing", () => {
-  let webext, helper;
+// Note: Update these selectors if the DOM changes in components
+const makeSelectors = ({ By }) => ({
```
lmorchard: Kind of figured these would change as UI work progresses. Could have followed the page object pattern, but forgot about it until just a little while ago. I can rework this if needed.

test/integration/functional-test.js
```diff
@@ -13,32 +13,261 @@ chai.use(chaiAsPromised);
...
+      await addLogin(mockLogin);
+      await waitUntilMissing(selectors.itemListEmpty);
+      await waitFor(selectors.itemList);
+
+      const listItemSubtitle = await waitFor(selectors.listItemSubtitle);
+      expect(await listItemSubtitle.getText()).to.equal(mockLogin.username);
+    });
+
+    it("displays login details on click", async () => {
+      // Add the login, then wait for "No results" to disappear.
```
lmorchard: This is a really hacky feeling sequence, but it seems to be the one that didn't end up with intermittent failures for me :/

6a68: I think this is a completely legit way of detecting DOM changes

PR #74, opened 2019-02-06
Disable prompts to save logins while on extension pages

### LINKED ISSUE #57: Disable login prompts in Management Interface

lmorchard:
- Add getLoginSavingEnabled and setLoginSavingEnabled methods to the
  Logins experiment API

- Disable login saving for extension pages on background startup

Fixes #57

### DIFF
--- a/src/background/environment.js
+++ b/src/background/environment.js
@@ -1,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export async function initializeEnvironment() {
+  // Issue #57: Disable login saving for extension pages.
+  const origin = browser.extension.getURL("").slice(0, -1);
+  const savingEnabled = await browser.experiments.logins
+    .getLoginSavingEnabled(origin);
+  if (savingEnabled) {
+    await browser.experiments.logins
+      .setLoginSavingEnabled(origin, false);
+  }
+}
--- a/src/experiments/logins/api.js
+++ b/src/experiments/logins/api.js
@@ -49,6 +49,20 @@ this.logins = class extends ExtensionAPI {
           //   - fooLogin = login as vanilla JS object
           //   - fooLoginInfo = login as nsILoginInfo
           //   - fooBag = (subset of) login fields as nsIPropertyBag
+          getLoginSavingEnabled(origin) {
+            try {
+              return Services.logins.getLoginSavingEnabled(origin);
+            } catch (ex) {
+              throw new ExtensionError(ex);
+            }
+          },
+          setLoginSavingEnabled(origin, isEnabled) {
+            try {
+              return Services.logins.setLoginSavingEnabled(origin, isEnabled);
+            } catch (ex) {
+              throw new ExtensionError(ex);
+            }
+          },
           getAll() {
             const logins = getLogins();
             return logins;
--- a/src/experiments/logins/schema.json
+++ b/src/experiments/logins/schema.json
@@ -200,6 +200,32 @@
     ],
     "functions": [
       {
+        "name": "getLoginSavingEnabled",
+        "type": "function",
+        "description": "Reports whether or not saving login information is enabled for a host.",
+        "async": true,
+        "parameters": [{
+          "name": "aHost",
+          "type": "string",
+          "description": "The origin for which to check the status of login saving"
+        }]
+      },
+      {
+        "name": "setLoginSavingEnabled",
+        "type": "function",
+        "description": "Enables or disables storing logins for a specified host. When login storing is disabled, the Login Manager won't prompt the user to store logins for that host. Existing logins are not affected.",
+        "async": true,
+        "parameters": [{
+          "name": "aHost",
+          "type": "string",
+          "description": "The origin for which to modify the login saving permission"
+        },{
+          "name": "isEnabled",
+          "type": "boolean",
+          "description": "The value to which to set login saving for the origin"
+        }]
+      },
+      {
         "name": "getAll",
         "type": "function",
         "description": "Gets an array of all Logins. Returns a Promise for the (possibly empty) list of all Logins.",
--- a/test/integration/logins-api-test.js
+++ b/test/integration/logins-api-test.js
@@ -46,6 +46,20 @@ describe("logins API", () => {
   let By;
   let until;
 
+  const enableRememberSignons = async () => {
+    await webext.inChrome();
+    await driver.executeScript(`
+      Services.prefs.setBoolPref("signon.rememberSignons", true);
+    `);
+  };
+
+  const disableRememberSignons = async () => {
+    await webext.inChrome();
+    await driver.executeScript(`
+      Services.prefs.setBoolPref("signon.rememberSignons", false);
+    `);
+  };
+
   const addLogin = async () => {
     await webext.inChrome();
     await driver.executeScript(`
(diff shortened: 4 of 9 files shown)


test/integration/logins-api-test.js
```diff
@@ -235,6 +249,45 @@ describe("logins API", () => {
...
+      it("enables login saving for origin when given true", async () => {
+        await commonTest(true, true);
+      });
+    });
+    describe("with pref signon.rememberSignons = false", () => {
+      before(async () => {
+        await disableRememberSignons();
+      });
+      it("does not enable login saving for origin when given true", async () => {
+        await commonTest(true, false);
```
lmorchard: Mild corner case here: If the user has turned off saving logins globally in options, `getLoginSavingEnabled` always returns false no matter what the setting might be.

src/background/environment.js
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export async function initializeEnvironment() {
+  // Issue #57: Disable login saving for extension pages.
+  const origin = browser.extension.getURL("").slice(0, -1);
+  const savingEnabled = await browser.experiments.logins
+    .getLoginSavingEnabled(origin);
+  if (savingEnabled) {
```
lmorchard: Checking whether saving is enabled to ensure we keep it off. The user could enable login save globally or manually delete the exception from the list in options. Kind of edge casey, but bleh.

lmorchard:
Hmm, so I only ensure saving is off on startup of the addon. The user could later enable saving in options or delete the exception - and then we'd see the login save prompt come up until next add-on startup. 

I wonder if I should make a check more often? Like, say, whenever the browser action button is clicked?

linuxwolf:
> Hmm, so I only ensure saving is off on startup of the addon. The user could later enable saving in options or delete the exception - and then we'd see the login save prompt come up until next add-on startup.
> 
> I wonder if I should make a check more often? Like, say, whenever the browser action button is clicked?

I think we can start with this, and see how it goes.  If we do it more frequently, I'd argue for whenever we think the management page is loaded.

src/background/environment.js
6a68: It seems unlikely that users are going to mess with this manually. That said, the browser APIs are available from extension pages, so we could just run this check when the management page is loaded (or, to get kind of extreme about the "user messes things up" edge case, we could run this check on page load and whenever the page visibility API's `visibilitychange` event fires).

src/experiments/logins/api.js
```diff
@@ -49,6 +49,20 @@ this.logins = class extends ExtensionAPI {
           //   - fooLogin = login as vanilla JS object
           //   - fooLoginInfo = login as nsILoginInfo
           //   - fooBag = (subset of) login fields as nsIPropertyBag
+          getLoginSavingEnabled(aHost) {
```
6a68: I really dislike the `aFoo` naming convention. How about just calling this `origin`?

src/experiments/logins/api.js
```diff
@@ -49,6 +49,20 @@ this.logins = class extends ExtensionAPI {
...
+            try {
+              return Services.logins.getLoginSavingEnabled(aHost);
```
6a68: The trailing slash thing seems like a super annoying gotcha. I think we should either handle it here (strip it out or throw if it's detected), or document this annoyance in the schema docs.

src/experiments/logins/schema.json
```diff
@@ -199,6 +199,32 @@
       }
     ],
     "functions": [
+      {
+        "name": "getLoginSavingEnabled",
+        "type": "function",
+        "description": "For the given origin, gets a boolean flag which determines whether login saving is enabled",
```
6a68: Should also mention that the returned promise is rejected if the URL is malformed.

src/experiments/logins/schema.json
```diff
@@ -199,6 +199,32 @@
...
+        "async": true,
+        "parameters": [{
+          "name": "aHost",
+          "type": "string",
+          "description": "The origin for which to check the status of login saving"
```
6a68: Just use the description from [the IDL](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsILoginManager#getLoginSavingEnabled()): `The hostname to check. This argument should be in the origin URL format, with no pathname. For example: "https://www.site.com".`

src/experiments/logins/schema.json
```diff
@@ -199,6 +199,32 @@
...
+        }]
+      },
+      {
+        "name": "setLoginSavingEnabled",
+        "type": "function",
+        "description": "For the given origin, uses the given value to set a boolean flag which determines whether login saving is enabled",
```
6a68: The IDL docs may be clearer here. Also, be sure to mention cases in which the returned promise is rejected, which I think is just if the URL is malformed.

src/experiments/logins/api.js
```diff
@@ -49,6 +49,20 @@ this.logins = class extends ExtensionAPI {
...
+            } catch (ex) {
+              throw new ExtensionError(ex);
+            }
+          },
+          setLoginSavingEnabled(aHost, isEnabled) {
```
6a68: How about `origin` and `shouldSaveLogins` for the arguments?

test/integration/logins-api-test.js
```diff
@@ -46,6 +46,20 @@ describe("logins API", () => {
   let By;
   let until;
 
+  const enableRememberSignons = async () => {
+    await webext.inChrome();
+    await driver.executeScript(`
+      Services.prefs.setBoolPref("signon.rememberSignons", true);
```
6a68: It would be good to actually call the Services.logins.get/setLoginSavingEnabled for a specific origin, then test it's correctly returned when you call the API. I wonder what happens if you accidentally leave in a trailing slash, for instance. This could be left for a followup commit.

test/integration/logins-api-test.js
6a68: Good catch! I'm not sure which, if any, parts of the logins API will work if logins are globally disabled. I think this needs investigation; could you file a bug?

src/experiments/logins/schema.json
6a68: Actually, looking at the schema docs, I haven't been super strict about reusing the IDL descriptions. This could be fine as is (but `aHost` does bug me ;-P).

6a68:
Whoops! Didn't see Matt had already reviewed. Sorry to pile on :-P

test/integration/logins-api-test.js
linuxwolf: From my own experiences, the CRUD portion of the upstream API works just fine regardless of what the global pref and this specific callout are set to.

test/integration/logins-api-test.js
linuxwolf: From my own experiences, the CRUD portion of the upstream API works just fine regardless of what the global pref and this specific callout are set to.

lmorchard:
I added a commit that automatically re-adds our entry to the exception list for login saving when it is somehow removed (manually or otherwise). However, this still needs unit / integration tests. However, however, this could use a sanity check as to whether this is what we actually want to do.

src/experiments/logins/api.js
lmorchard: Heh, I used aHost to match the upstream API, but I can change it if necessary

src/background/environment.js
```diff
@@ -0,0 +1,25 @@
...
+  // If login saving is enabled for us on startup, add an exception.
+  const savingEnabled = await browser.experiments.logins
+    .getLoginSavingEnabled(ourOrigin);
+  if (savingEnabled) {
+    await browser.experiments.logins
+      .setLoginSavingEnabled(ourOrigin, false);
+  }
+
+  // HACK: If our exception to login saving is removed, immediately re-add it.
+  browser.experiments.logins.onLoginSavingChanged
```
lmorchard: Okay, here's the bit that could use some sanity checking. I worked out how to subscribe to the event that fires when a login saving exception is removed, and automatically re-add it. This may be a mistake, but I can't think of a reason why

src/experiments/logins/api.js
lmorchard: Yeah, same as above, I was trying to follow the underlying API rather than invent new names

lmorchard:
Actually, you know what: Never mind. It occurred to me over the weekend that this PR *without* the event handler to automatically re-enable the exception was already 80% of the solution and pretty much reviewed.

My commit that uses the permission events still needs time to bake, so I'll do it in a follow-up PR.

punkbusterdani:
I wish to verify the fix of this issue, but I'm lacking some steps here. How am I supposed to reproduce/verify this issue? Thanks.

PR #80, opened 2019-02-13
Expect a complete Login for update in Login API

### LINKED ISSUE #58: Logins API - require complete Login for update
Followup issue from [last bit of review feedback](https://github.com/mozilla-lockbox/lockbox-addon/pull/47#pullrequestreview-192706763).

The update() method should require a complete login object, rather than just the bits we've updated, so that we don't have to internally strip out `null` fields injected by the webextensions framework

lmorchard:
Fixes #58

## Testing and Review Notes

Nothing apparent should change. Integration tests should pass, manual item editing should work as expected.

### DIFF
--- a/src/experiments/logins/api.js
+++ b/src/experiments/logins/api.js
@@ -84,32 +84,19 @@ this.logins = class extends ExtensionAPI {
               throw new ExtensionError(ex);
             }
           },
-          update(newLogin) {
-            if (!("guid" in newLogin)) {
+          update(updatedLogin) {
+            if (!("guid" in updatedLogin)) {
               throw new ExtensionError(`Update failed: ID is required`);
             }
-            // The addons framework code always supplies null values for missing keys,
-            // so filter those out.
-            // TODO: Could this cause subtle bugs? Would anyone ever want to update a field to be null? If so, maybe default to 'NULL" (the string), or undefined.
-            for (let key in newLogin) {
-              if (newLogin[key] === null) {
-                delete newLogin[key];
-              }
-            }
-            const oldLogin = getLogin(newLogin.guid);
+            const oldLogin = getLogin(updatedLogin.guid);
             if (!oldLogin) {
-              throw new ExtensionError(`Update failed: Login not found with ID ${newLogin.guid}`);
+              throw new ExtensionError(`Update failed: Login not found with ID ${updatedLogin.guid}`);
             }
             try {
-              const updatedLogin = {};
-              Object.assign(updatedLogin, oldLogin, newLogin);
-              if ("timesUsedIncrement" in newLogin) {
-                delete updatedLogin.timesUsed;
-              }
-              const updatedLoginBag = LoginHelper.newPropertyBag(updatedLogin);
+              const updatedLoginInfo = LoginHelper.vanillaObjectToLogin(updatedLogin);
               const oldLoginInfo = LoginHelper.vanillaObjectToLogin(oldLogin);
-              Services.logins.modifyLogin(oldLoginInfo, updatedLoginBag);
-              const finalLogin = getLogin(newLogin.guid);
+              Services.logins.modifyLogin(oldLoginInfo, updatedLoginInfo);
+              const finalLogin = getLogin(updatedLogin.guid);
               return finalLogin;
             } catch (ex) {
               throw new ExtensionError(ex);
--- a/test/integration/test-pages/logins-api.js
+++ b/test/integration/test-pages/logins-api.js
@@ -43,12 +43,10 @@ document.querySelector("#remove").addEventListener("click", () => {
 });
 document.querySelector("#update").addEventListener("click", () => {
   const log = getLogger("update");
-  const loginUpdates = {
-    guid: mockLogin.guid,
-    username: "updated",
-  };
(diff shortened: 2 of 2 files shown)


src/experiments/logins/api.js
```diff
@@ -84,32 +84,19 @@ this.logins = class extends ExtensionAPI {
...
+              throw new ExtensionError(`Update failed: Login not found with ID ${updatedLogin.guid}`);
             }
             try {
-              const updatedLogin = {};
-              Object.assign(updatedLogin, oldLogin, newLogin);
-              if ("timesUsedIncrement" in newLogin) {
-                delete updatedLogin.timesUsed;
-              }
-              const updatedLoginBag = LoginHelper.newPropertyBag(updatedLogin);
+              const updatedLoginInfo = LoginHelper.vanillaObjectToLogin(updatedLogin);
```
lmorchard: I scratched my head at the description of this issue for a little bit. But, I think the crux of the issue is where [modifyLogin()](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsILoginManager#modifyLogin()) can accept either an nsIPropertyBag or an nsILoginMetaInfo instance.

Switched to the latter, here, on assumption of a full login object modified by the calling code as necessary, rather than a sparse set of updates. So I *think* that's me understanding the issue correctly.

test/integration/test-pages/logins-api.js
```diff
@@ -43,12 +43,10 @@ document.querySelector("#remove").addEventListener("click", () => {
...
   const log = getLogger("update");
-  const loginUpdates = {
-    guid: mockLogin.guid,
-    username: "updated",
-  };
-  browser.experiments.logins.update(loginUpdates).
-    then(log, log);
+  browser.experiments.logins.get(mockLogin.guid)
+    .then(login =>
+      browser.experiments.logins.update({ ...login, username: "updated" }))
```
lmorchard: Oh, and also I _think_ this represents correctly the change in assumptions. I was already using the API this way in datastore.js, oddly enough, which made no functional difference really.

src/experiments/logins/api.js
6a68: I think it'd be better to stick with the PropertyBag as the second argument to `modifyLogin`.

Sorry this wasn't clearer from the bug description. The LoginInfo fields can be updated in this way, but not the [LoginMetaInfo](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsILoginMetaInfo) fields (timestamps and usage count). As I understand it, the only reason there's a separate interface for the MetaInfo fields is that they were needed after login sync was already done, so each login object actually implements both interfaces: the syncable part, and the non-syncable part. Even with the historical background in mind, it's a confusing and, uh, suboptimal situation.

We could just rely on the `touch()` method to alter the MetaInfo fields, but that makes `update()` an annoyingly misleading function name, and causes the grody dual-interface situation to complicate our lives.

test/integration/test-pages/logins-api.js
6a68: This looks good, but I'm wondering if we should add clues about the special MetaInfo fields (including the magic [`timesUsedIncrement`](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsILoginManager#modifyLogin()) property). Would you mind also setting `timesUsedIncrement: 1` here, and also bump one of the MetaInfo timestamp fields, then adding the corresponding asserts on the test side? Maybe this'll help future contributors detect the weirdness of the underlying APIs.

lmorchard: Banged my head on this for awhile and I think I've discovered this: 

If we supply both `timesUsed` and `timesUsedIncrement` in the property bag at the same time, the explicit `timesUsed` value will likely clobber whatever the result of `timesUsedIncrement` might be. 

That is, [depending on whatever the order of enumerated properties ends up being](https://searchfox.org/mozilla-central/source/toolkit/components/passwordmgr/LoginHelper.jsm#363). I haven't dug deep enough to figure out what that order is, though it seemed pretty consistently to clobber the increment.

On the bright side, I was able to update `timePasswordChanged`.

src/experiments/logins/api.js
lmorchard: Switched back to a property bag. But, it looks like we might have to rely on `touch()` for using the `timesUsedIncrement` property.

linuxwolf: relying on `touch()` seems reasonable to me.

test/integration/test-pages/logins-api.js
linuxwolf: I think it's reasonable for `update()` to actively omit `timesUsedIncrement` -- either the user of this API is properly incrementing `timesUsed` by hand, or they can use `touch()`.

thoughts? @lmorchard @6a68

lmorchard:
Okie doke, I'll merge this and then we can see if anything else needs doing from here.

lmorchard:
FWIW, I was about to file a follow-up issue since I thought `timePasswordChanged` wasn't updated automatically - [but it looks like that *does* happen](https://searchfox.org/mozilla-central/source/toolkit/components/passwordmgr/LoginHelper.jsm#329).

Issue #29, opened 2018-12-14
Add telemetry events for relevant actions in the extension


sandysage:
As the product owner, I want to know how users engage with the app to help inform decisions for future effort, while making sure users have control over whether or not their usage data is submitted.

## Acceptance criteria

* All events should send Telemetry event data defined by the Mozilla product, data science, and legal teams
* The user should be able to disable usage data collection (Telemetry) via a setting

sandysage:
I think this probably starts as modifying [the existing metrics.md](https://github.com/mozilla-lockbox/lockbox-addon/blob/master/docs/metrics.md).

6a68:
Implementation will be straightforward: the telemetry API exposes everything we need. Most of the work in this bug will be the legal & data reviews.

6a68:
> Most of the work in this bug will be the legal & data reviews.

🤣

Issue #5, opened 2018-11-29
Use Telemetry WebExtension API (privileged)


sandysage:
As the product owner, I want to know how users engage with the extension to help inform decisions for future effort, while making sure users have control over whether or not their usage data is submitted.

## Acceptance criteria
* All events should send Telemetry event data defined by the Mozilla product, data science, and legal teams
* The user should be able to disable usage data collection (Telemetry) via a setting
* Should use the new Telemetry API as it's necessary to function post Firefox 64

linuxwolf:
This _should_ be done as part of migrating to this repo, but need to confirm.

6a68:
This was done as part of #29

PR #82, opened 2019-02-19
edit metrics doc

irrationalagent:
Connected to https://github.com/mozilla-lockbox/lockbox-addon/issues/29 and https://github.com/mozilla-lockbox/lockbox-addon/issues/5

## Testing and Review Notes

These are the metrics I'm proposing for the first iteration of the addon. 

I based the event schemas on the [invision designs](https://mozilla.invisionapp.com/share/ENOL84959AK#/screens/339614512). If there's other nomenclature that is better compatible with what's in the code, I'm fine with that.

Also, note that a couple of scalars that I proposed might pose technical hurdles. Happy to talk about that. 

I also left out the monitor bits for now.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -79,7 +79,7 @@ Services.telemetry.registerEvents("event_category", {
 For our purposes, we will use the `extra` field for a few purposes:
 
 - To log the UUID of the item that has been added or changed (e.g. `"item_id": UUID`)
-- To log the fields that are modified when an item is updated in the datastore (e.g. `"fields": "username,passsword,address"` (because the value has to be a string we will have to concat the fields that were updated somehow)
+- To log the fields that are modified when an item is updated in the datastore (e.g. `"fields": "username,passsword,hostname"` (because the value has to be a string we will have to concat the fields that were updated somehow)
 
 Once an event is registered, we can record it with:
 
@@ -131,9 +131,9 @@ All events are currently implemented under the **category: lockbox-addon**. The 
 
 1. `startup` fires when the webextension is loaded. **objects**: webextension. Note that this event fires whenever the browser is started, so is not indicative of direct user interaction with Lockbox. **value** is null.
 
-2. `click` fires when someone clicks on one of the UI elements listed **objects**: `toolbar`, `get_mobile`, `faq`, `account_settings`, `give_feedback`, `settings_menu` (or whatever the menu is that contains links to account settings, faq, etc). **value** is null.
+2. `click` fires when someone clicks on one of the UI elements listed **objects**: `toolbar`, `get_mobile`, `faq`, `account_settings`, `give_feedback`, `settings_menu`, `reconnect_sync`, `signin_sync` (or whatever the menu is that contains links to account settings, faq, etc). **value** is null.
 
-3. `show` events fire when various UI elements are rendered shown to the user. **objects**: `item_list_manager`, `item_list_doorhanger`, `item_detail_doorhanger` `item_detail_manager`, `new_item`, `item_edit`, `delete_confirm`, `connect_another_device`, `reconnect_sync` **value** should be a boolean for `item_list_*` events indicating whether any logins are in the login list (e.g. `False` if the list is empty).
+3. `show` events fire when various UI elements are rendered shown to the user. **objects**: `item_list_manager`, `item_list_doorhanger`, `item_detail_doorhanger` `item_detail_manager`, `new_item`, `item_edit`, `delete_confirm`, `connect_another_device` (referring to the dialog displayed after a user clicks on the button to learn about the mobile apps) **value** should be a boolean for `item_list_*` events indicating whether any logins are in the login list (e.g. `False` if the list is empty).
 
 4. `item_add`, `item_update`, `item_delete` fire after a successful adding, editing or deleting of a credential. **objects**: manager, doorhanger (contingent on where the user initiated the action). item GUID should be in the extra field. **value** is null
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -62,13 +62,13 @@ The events that we will record will conform to the structure expected by the tel
 The API takes care of the timestamp for us. We just need to define `category`, `method`, `object` and `extra` (`value` is optional and we won't use it).
 
 
-Events are defined by **registering** them using a call to `Services.telemetry.registerEvents(category, eventData)`.
+Events are defined by **registering** them using a call to `browser.telemetry.registerEvents(category, eventData)`.
 
 Here's a breakdown of how a to register a typical event:
 
 
 ```javascript
-Services.telemetry.registerEvents("event_category", {
+browser.telemetry.registerEvents("event_category", {
     "event_name": {
         methods: ["click", ... ], // types of events that can occur
         objects: ["a_button", ... ], // objects event can occur on
@@ -83,7 +83,7 @@ For our purposes, we will use the `extra` field for a few purposes:
 
 Once an event is registered, we can record it with:
 
-`Services.telemetry.recordEvent(category, method, object, value, extra)`
+`browser.telemetry.recordEvent(category, method, object, value, extra)`
 
 Note: The semantics of `value` is contingent on the event being recorded, see list below.
 
@@ -94,9 +94,9 @@ See the Events section for specific examples of event registration and recording
 We use the js api for scalar recording as well. Here registration happens with the following syntax:
 
 ```javascript
-Services.telemetry.registerScalars(category, {
+browser.telemetry.registerScalars(category, {
 	"scalar_name": {
-		kind: services.Telemetry.SCALAR_TYPE_COUNT, // SCALAR_TYPE_COUNT, SCALAR_TYPE_BOOLEAN. or SCALAR_TYPE_STRING
+		kind: browser.Telemetry.SCALAR_TYPE_COUNT, // SCALAR_TYPE_COUNT, SCALAR_TYPE_BOOLEAN. or SCALAR_TYPE_STRING
 		keyed: false,
 		record_on_release: false, // NEEDS TO BE SET TO RECORD ON RELEASE CHANNEL
 		expired: false,
@@ -105,7 +105,7 @@ Services.telemetry.registerScalars(category, {
 We set scalar values in the following way:
 
 ```javascript
-Services.telemetry.scalarSet(
+browser.telemetry.scalarSet(
 	"category.scalar_name", value
 );
 ```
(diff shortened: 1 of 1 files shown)


docs/metrics.md
```diff
@@ -124,39 +109,44 @@ Services.telemetry.scalarSet(
...
 We can also use `scalarAdd` to increment a scalar value by some amount.
 
 ## Scalar Metrics
 
-These are the metrics we currently collect regarding the state of user datastores.
+These are the metrics we currently collect regarding the state of the user's login storage.
+
+- `loginCount` (SCALAR_TYPE_COUNT). Current count of the number of items in the user's login storage. If possible, this scalar should be set upon opening either the doorhanger or full extension interface.
+
+- `loginsAccessedAllTime` (SCALAR_TYPE_COUNT). Running total of logins accessed (copied, revealed) through the extension interface or doorhanger. Should never be reset.
```
linuxwolf: it's not clear to me, from the interface descriptions, how possible either this or `loginsAccessed` is from the Telemetry WebExtension API.  One of them definitely is, but it's not clear which; not clear how much effort it will take to have both.

linuxwolf:
Would still very much want to get @6a68 thoughts.

docs/metrics.md
```diff
@@ -94,13 +79,13 @@ Services.telemetry.registerEvents("event_category", {
 For our purposes, we will use the `extra` field for a few purposes:
 
 - To log the UUID of the item that has been added or changed (e.g. `"item_id": UUID`)
-- To log the fields that are modified when an item is updated in the datastore (e.g. `"fields": "password,notes"`  (because the value has to be a string we will have to concat the fields that were updated somehow)
+- To log the fields that are modified when an item is updated in the datastore (e.g. `"fields": "username,passsword,address"` (because the value has to be a string we will have to concat the fields that were updated somehow)
```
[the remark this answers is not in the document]
irrationalagent: yes, I should change it to `hostname`

docs/metrics.md
```diff
@@ -124,39 +109,44 @@ Services.telemetry.scalarSet(
...
+e.g. `lockbox-addon.loginCount` for the scalar name.
 
 We can also use `scalarAdd` to increment a scalar value by some amount.
 
 ## Scalar Metrics
 
-These are the metrics we currently collect regarding the state of user datastores.
+These are the metrics we currently collect regarding the state of the user's login storage.
+
+- `loginCount` (SCALAR_TYPE_COUNT). Current count of the number of items in the user's login storage. If possible, this scalar should be set upon opening either the doorhanger or full extension interface.
```
6a68: We can definitely trigger this ping when the user opens the doorhanger or loads the management page.

Would it also be useful to check this value if the user doesn't interact with Lockbox? For instance, we could check this value at browser startup, or periodically (whenever Firefox is open).

docs/metrics.md
```diff
@@ -1,71 +1,57 @@
...
 
 ## Collection
 
-At this point, all measurements related to Lockbox will be made client-side. However, future releases will give users the option to sync their Lockbox data via an FxA account, at which point additional measurements will be logged server-side through the FxA data pipeline. We are not directly responsible for the measurements made through that mechanism.
-
-For our internal alpha release, we will be making use of the public JavaScript API that allows recording and sending of event data and scalar data through an add-on. The API is documented here:
-
-https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#the-api
+To record Telemetry we will be making use of the public JavaScript API that allows recording and sending of [events](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#public-js-api) and [scalar measurements](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/scalars.html#js-api) through a webextension.
```
6a68: About the use of the term 'public' here: the telemetry API is only exposed to Mozilla addons; 'Mozilla privileged' might be more accurate.

docs/metrics.md
6a68: I'm pretty sure that `scalarAdd` increments a per-browsing session counter, so `loginsAccessedAllTime` would require saving the all-time count in webextension browser storage (not that hard to do).

docs/metrics.md
```diff
@@ -124,39 +109,44 @@ Services.telemetry.scalarSet(
...
 ## Scalar Metrics
 
-These are the metrics we currently collect regarding the state of user datastores.
+These are the metrics we currently collect regarding the state of the user's login storage.
+
+- `loginCount` (SCALAR_TYPE_COUNT). Current count of the number of items in the user's login storage. If possible, this scalar should be set upon opening either the doorhanger or full extension interface.
+
+- `loginsAccessedAllTime` (SCALAR_TYPE_COUNT). Running total of logins accessed (copied, revealed) through the extension interface or doorhanger. Should never be reset.
+
+- `loginsAccessed` (SCALAR_TYPE_COUNT). Total of logins accessed (copied, revealed) since last ping through the extension interface or doorhanger. TODO: in native code scalars are automatically reset when a new ping is cut. Will that happen for us?
```
6a68: This should be easy on a per-session basis.

I believe the pings should be reset the same way as native code, since the `browser.telemetry` APIs are just a passthrough to the `Services.telemetry` APIs. @georgf might know for sure.

docs/metrics.md
```diff
@@ -124,39 +109,44 @@ Services.telemetry.scalarSet(
...
 
-2. `iconClick` fires when someone clicks the toolbar icon. **objects**: toolbar
+3. `show` events fire when various UI elements are rendered shown to the user. **objects**: `item_list_manager`, `item_list_doorhanger`, `item_detail_doorhanger` `item_detail_manager`, `new_item`, `item_edit`, `delete_confirm`, `connect_another_device` (referring to the dialog displayed after a user clicks on the button to learn about the mobile apps) **value** should be a boolean for `item_list_*` events indicating whether any logins are in the login list (e.g. `False` if the list is empty).
 
-3. `render` events fire when the item manager or doorhanger (when implemented) are rendered. **objects**: manage, doorhanger
+4. `item_add`, `item_update`, `item_delete` fire after a successful adding, editing or deleting of a credential. **objects**: manager, doorhanger (contingent on where the user initiated the action). item GUID should be in the extra field. **value** is null
```
6a68: Item GUID might be a privacy concern, since it's unique to a specific login owned by a specific user. Would we want to hash this value before submitting?

docs/metrics.md
```diff
@@ -124,39 +109,44 @@ Services.telemetry.scalarSet(
...
 
-All events are currently implemented under the **category: lockboxV2**. The `extra` field contains `itemid` for events pertaining to a particular Lockbox item. They are listed and grouped together below based on the contents of the event's `method` field.
+1. `startup` fires when the webextension is loaded. **objects**: webextension. Note that this event fires whenever the browser is started, so is not indicative of direct user interaction with Lockbox. **value** is null.
 
-1. `startup` fires when the webextension is loaded. **objects**: webextension. Note that this event fires whenever the browser is started, so is not indicative of direct user interaction with Lockbox.
+2. `click` fires when someone clicks on one of the UI elements listed **objects**: `toolbar`, `get_mobile`, `faq`, `account_settings`, `give_feedback`, `settings_menu`, `reconnect_sync`, `signin_sync` (or whatever the menu is that contains links to account settings, faq, etc). **value** is null.
```
6a68: Do we want a separate event if users do things with the keyboard instead of the mouse? (Accessibility and keyboard-centric power users)

docs/metrics.md
georgf: That’s correct, they reset per ping/subsession.

The semantics for the web extension Telemetry API are the same, it’s just a different way to access them.

docs/metrics.md
linuxwolf: Relaying from an earlier IRL conversation when I asked exactly this: these metrics are also measuring as user's engagement with Lockbox specifically, so it's preferable to limit to triggers specific to interacting with Lockbox.

docs/metrics.md
linuxwolf: I'm curious about the concern.  The value itself is randomly generated (not based on or derived from the login information) from a generic entropy pool (read: not based or derived from the user).  The events are not carrying any other information about the login (e.g., username, hostname, etc) or into what the interaction is concerned (e.g., the webpage's origin).

For some of the measures in this space to have meaning, we'd have to consistently hash the GUID across platforms and instances.

docs/metrics.md
irrationalagent: Yes +1 to Matt, trying to limit pings with extension data to those who are actually using it (makes counting things like "active" users a bit easier).

docs/metrics.md
irrationalagent: Thanks for the clarification Georg. I'm going to remove the All-time scalar for now, this is relatively straightforward to compute on a per-client basis. It also mitigates some of the anonymity issues discussed with Jared.

docs/metrics.md
irrationalagent: For clarification on why I was asking for this, the GUID allows me to answer, e.g., how many UNIQUE entries were edited per user. I also realized that I left this out for the credential access actions, so I added it for those as well.

If there is an alternative way, e.g. set a scalar that's like `uniqueLoginsAccessed` then we could dispense with adding the GUID.

docs/metrics.md
irrationalagent: Good question. I don't see myself utilizing that information, but if it would help engineering to have that data I'm happy to add it. And I should say more broadly, if there is any data that engineering would find useful to collect, we should add that.

docs/metrics.md
linuxwolf: > If there is an alternative way, e.g. set a scalar that's like `uniqueLoginsAccessed` then we could dispense with adding the GUID.

Knowing how users are interacting with their logins has been helpful to prioritize work, so the granularity wherein is helpful if we can keep it.  I'm not sure how reliably we could track a scalar, though, especially across sessions.

irrationalagent: OK. To be clear, the fact that scalars naturally reset on telemetry sessions isn't a concern for me anymore.

6a68: > I'm curious about the concern. The value itself is randomly generated (not based on or derived from the login information) from a generic entropy pool (read: not based or derived from the user). The events are not carrying any other information about the login (e.g., username, hostname, etc) or into what the interaction is concerned (e.g., the webpage's origin).

Good points. Maybe I'm being excessively cautious.

docs/metrics.md
6a68: I guess it's really a question of whether product would find the key vs. click (vs. tap, since some desktops have touch screens...) distinction useful or not.

linuxwolf: @mozilla-lockbox/product is the distinction (click vs tap vs keypress) useful to track?

sandysage: It'd be cool to track but I don't think it's going to give us relevant insights on their own. For now, we probably want to just track that someone did the thing. We can always layer in the distinction at a later time when we have a specific need to identify the methods by which people navigate.

6a68:
Rebased and force-pushed.

PR #84, opened 2019-02-21
"Reveal Password" functionality

### LINKED ISSUE #20: "Reveal Password" functionality
When I want to access an account on a different device, or copy/paste doesn’t work for me, I want to see the password so I can manually enter it to get into my account.

## Acceptance criteria

* Should be able to view password from detail view

## Visual Design

**InVision**

**Zeplin**
Password Hidden (default): https://zpl.io/boK9ggk
Password Shown: https://zpl.io/bzARAzX
Password - Long (read version): https://zpl.io/aMKJOzK
Password - Long (edit mode): https://zpl.io/adrvOB7 (field would be scrollable)

lmorchard:
- New PasswordText widget with hide/show buttons

- Tweak hide/show and copy button alignments

- Smaller PASSWORD_DOT

- Replace hide / show eye icons

Fixes #20

## Testing and Review Notes

- Viewing an entry should now allow show / hide of password and not just simple dots
- Show / hide button should no longer shift position on editing view

## To Do

- [ ] "long password" case does not grow wider in item view, exploring a solution


src/widgets/password-text.js
```diff
@@ -0,0 +1,73 @@
...
+
+import { Localized } from "fluent-react";
+import PropTypes from "prop-types";
+import React from "react";
+
+import { classNames } from "../common";
+
+import styles from "./input.css";
+
+const PASSWORD_DOT = "\u2022";
```
lmorchard: This unicode char "•" more closely matches the Zeplin mock - but the previous value "●" more closely matches the built-in password field dots. This is easy to switch back if we want.

src/widgets/password-text.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  render() {
+    const {className, value, ...props} = this.props;
+    const {showPassword} = this.state;
+
+    return (
+      <div {...props} className={classNames([
+        styles.monospace,
+        styles.passwordText,
```
lmorchard: Two issues I ran into here, trying to handle [a "long password" case](https://app.zeplin.io/project/5c1829b8fde4d550a26b270f/screen/5c3d0105a8ecedbdc31cdb22) here:
1. How long is long? (e.g. 15 characters?) We could possibly switch a classname on value length
2. The whole form is on a shared CSS grid, with the copy buttons aligned. Breaking that alignment calls for a different overall layout for the form, I think? My CSS-fu fails me here.

src/widgets/input.css
```diff
@@ -103,9 +126,18 @@ input::placeholder {
...
+  right: 0;
+  background: rgba(255, 255, 255, 1.0);
+}
+
+.password-text button::before,
 .password button::before {
+  cursor: pointer;
   background-repeat: no-repeat;
   background-size: 16px 16px;
+  background-position: 0 2px;
```
lmorchard: This is kind of arbitrary, but seems to get the show/hide icon better aligned with the dots and midpoint of text. Might be worth yanking out.

src/widgets/input.css
```diff
@@ -72,7 +74,28 @@ input::placeholder {
   border: 0;
   color: #0060df;
   background-color: transparent;
-  padding: 4px 8px;
+  padding: 4px 8px 4px 12px;
+}
+
+.password-text {
+  width: 136px;
```
lmorchard: I copied a lot of these properties straight from Zeplin - some of it seems redundant and could probably get scrubbed out.

src/icons/hide.svg
```diff
@@ -1,7 +1,16 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
-  <g fill="none" fill-rule="evenodd" transform="translate(0 1)">
-    <path stroke="#0C0C0D" stroke-opacity=".6" stroke-width="2" d="M1,6.95041322 C1,9.20254695 4.73761678,11.9008264 8,11.9008264 C11.2623832,11.9008264 15,9.20254695 15,6.95041322 C15,4.6982795 11.2623832,2 8,2 C4.73761678,2 1,4.6982795 1,6.95041322 Z"/>
-    <path fill="#6D6D6E" d="M14.0160542,0.675736079 C14.3979633,1.05764518 14.3979633,1.67684218 14.0160542,2.05875128 L2.95193261,13.1228729 C2.5700235,13.504782 1.95082651,13.504782 1.56891741,13.1228729 C1.1870083,12.7409638 1.1870083,12.1217668 1.56891741,11.7398577 L12.633039,0.675736079 C13.0149481,0.293826977 13.6341451,0.293826977 14.0160542,0.675736079 Z"/>
-    <path fill="#6D6D6E" d="M9.60800171,4.46955872 C9.1423119,4.17216421 8.59352469,4 8,4 C6.34314575,4 5,5.34314575 5,7 C5,7.5968592 5.1629633,8.14651554 5.46343994,8.61383057 L9.60800171,4.46955872 Z"/>
-  </g>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="16" height="16" viewBox="0 0 16 16">
```
lmorchard: Could probably run some optimization on these SVGs, but I just copied them out of the Zeplin mock

lmorchard:
Doh. Broke a handful of unit tests - handling that right now.

src/widgets/copy-to-clipboard-button.js
```diff
@@ -65,19 +63,18 @@ export default class CopyToClipboardButton extends React.Component {
         </Localized>
       );
     }
-    return (
-      <Stack stretch selectedIndex={selectedIndex} className={className}>
```
lmorchard: Oh yeah, and I meant to mention in my earlier comments that I found some of the CSS stuff a bit easier after dropping the Stack component (which was the main cause of the failed unit tests, so that reminded me)

src/widgets/password-text.js
linuxwolf: @mozilla-lockbox/ux some insights here would be helpful

>     1. How long is long? (e.g. 15 characters?) We could possibly switch a classname on value length

I would start with 15 or 20 personally, too

>     2. The whole form is on a shared CSS grid, with the copy buttons aligned. Breaking that alignment calls for a different overall layout for the form, I think? My CSS-fu fails me here.

I think the intent of the designs is for the reveal button to be adjacent to the password text, which means it's position could shift based on how long the value is.

To keep the reveal in a fixed spot, I think you're right that it would require changing the CSS grid column count.

lmorchard: Well, I've got the reveal button sitting at the end of the text, which seems to match the spec. It's the unaligned copy button and expanding the width of the field that seems a problem

nickbrandt:
After a discussion with @lmorchard regarding following our grid layout with the copy buttons, we have decided to move forward with the copy actions all staying aligned, even when one is longer than the rest. The designs  have been updated to reflect this.

src/widgets/password-text.js
linuxwolf: sorry, I was misunderstanding completely!  And it looks like the UX questions are worked out.

PR #96, opened 2019-03-06
Add sorting to lists

### LINKED ISSUE #35: Sort entries on management interface
When I have several Lockbox entries and I need to access an online account, I need to be able to quickly find an entry by website address, entry name, or username.

## Acceptance criteria
* Should sort alphabetically (default) or by the last used and last changed
* Should not sort on any additional meta data which includes # of times used and date created
* Should only be enabled on the management interface and not the doorhanger to prevent user confusion

## Visual Design
**Zeplin**
https://zpl.io/bWgLA9M

6a68:
Cleaned up and ready to land.

Fixes #35, #93.

## Testing and Review Notes

Adds management page UI to sort the list by name, last changed, or last used. Sorting should work while filtering. Sort choice should persist across manage page reloads.

Popup entries should always be displayed in last used order.

## Screenshots or Videos

_(Optional: to clearly demonstrate the feature or fix to help with testing and reviews)_

## To Do

- add “WIP” to the PR title if pushing up but not complete nor ready for review
- [x] add unit tests

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -30,6 +30,9 @@ module.exports = (config) => {
         { type: "html", subdir: "html" },
       ],
     },
+    mochaReporter: {
+      showDiff: true,
+    },
 
     webpack: require("./webpack.config"),
     webpackMiddleware: {
--- a/src/icons/arrow-dropdown.svg
+++ b/src/icons/arrow-dropdown.svg
@@ -1,0 +1,1 @@
+<?xml version="1.0" ?><svg height="24" viewBox="0 0 48 48" width="24" xmlns="http://www.w3.org/2000/svg"><path d="M14 20l10 10 10-10z"/><path d="M0 0h48v48h-48z" fill="none"/></svg>
--- a/src/list/manage/components/item-list-panel.css
+++ b/src/list/manage/components/item-list-panel.css
@@ -2,17 +2,48 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-.filter-toolbar > * + * {
-  margin-inline-start: 8px;
+.filter-toolbar {
+  flex-direction: column;
+  align-items: stretch;
 }
 
 .panel-header {
   composes: panel-header from "../../../widgets/panel.css";
   background: #fff;
+  height: auto;
+}
+
+.first-row, .second-row {
+  display: flex;
+  flex-flow: row nowrap;
+}
+
+.first-row > * + * {
+  margin-inline-start: 8px;
+}
+
+.second-row {
+  font-weight: normal;
+  display: flex;
+  justify-content: space-between;
+  margin-inline-start: 0;
+  line-height: 32px;
+  height: 32px;
+  margin-bottom: -12px;
+  margin-top: 8px;
 }
 
 .item-filter {
   composes: item-filter from "../../../widgets/panel.css";
   background: #f3f3f5;
   height: 32px;
+}
+
+.flex-spacer {
+  flex-grow: 1;
+}
+
+.list-counter {
+  flex-grow: 0;
+  flex-shrink: 0;
 }
--- a/src/list/manage/components/item-list-panel.js
+++ b/src/list/manage/components/item-list-panel.js
@@ -10,11 +10,13 @@ import { PanelHeader, PanelBody } from "../../../widgets/panel";
 import AddItem from "../containers/add-item";
 import ItemList, { ItemListPlaceholder } from "../../components/item-list";
 import ItemFilter from "../../containers/item-filter";
+import ListSort from "../containers/list-sort.js";
+import ListCounter from "../components/list-counter.js";
 
 import styles from "./item-list-panel.css";
 
 export default function ItemListPanel({className, inputRef, totalItemCount,
-                                       ...props}) {
+                                       sort, count, ...props}) {
   const hasItems = props.items.length !== 0;
   let list;
   if (!hasItems) {
@@ -35,9 +37,16 @@ export default function ItemListPanel({className, inputRef, totalItemCount,
       <PanelHeader className={styles.panelHeader}
                    border={hasItems ? "floating" : "none"}
                    toolbarClassName={styles.filterToolbar}>
-        <ItemFilter className={styles.itemFilter}
-                    inputRef={inputRef}/>
-        <AddItem/>
+        <div className={styles.firstRow}>
+          <ItemFilter className={styles.itemFilter}
+                      inputRef={inputRef}/>
+          <AddItem/>
+        </div>
+        <div className={styles.secondRow}>
+          <ListSort sort={sort} {...props} />
+          <div className={styles.flexSpacer}></div>
+          <ListCounter count={count} className={styles.listCounter} />
+        </div>
       </PanelHeader>
 
       <PanelBody scroll={false}>
--- a/src/list/manage/components/list-sort.js
+++ b/src/list/manage/components/list-sort.js
@@ -1,0 +1,68 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Localized } from "fluent-react";
+import PropTypes from "prop-types";
+import React from "react";
+
+import styles from "./list-sort.css";
+
+export default class ListSort extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: props.value,
+    };
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(evt) {
+    const value = evt.target.value;
+    this.setState({ value });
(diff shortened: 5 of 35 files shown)


src/list/manage/components/item-list-panel.js
```diff
@@ -38,6 +39,7 @@ export default function ItemListPanel({className, inputRef, totalItemCount,
         <ItemFilter className={styles.itemFilter}
                     inputRef={inputRef}/>
         <AddItem/>
+        <ListSort sort={sort} className={styles.listSort} {...props} />
```
linuxwolf: This layout puts everything into a single row.  Maybe we should consider some options:
* moving `<ListSort/>` (or a wrapper of it?) up a level to be a sibling of `<PanelHeader/>` and `<PanelBody/>`; or
* making `<PanelHeader/>`'s orientation configurable (flex column vs flex row)

I fear trying to make `<PanelHeader/>` a grid is more trouble than it's worth ...

src/widgets/select.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  focus() {
+    this.selectElement.focus();
+  }
+
+  render() {
+    const { className, options, disabled, ...props } = this.props;
+    return (
+      <select value={this.state.value}
```
linuxwolf: From past experience, trying to style `<select/>` to the degree our designs call for has been very problematic (maybe impossible).  I think this `<Select/>` component is a good abstraction over what we might ultimately do, though it's worth taking a bit to see if such a component exists out in the larger ecosystem.

src/list/manage/components/item-list-panel.js
```diff
@@ -35,9 +37,16 @@ export default function ItemListPanel({className, inputRef, totalItemCount,
...
-        <ItemFilter className={styles.itemFilter}
-                    inputRef={inputRef}/>
-        <AddItem/>
+        <div className={styles.firstRow}>
+          <ItemFilter className={styles.itemFilter}
+                      inputRef={inputRef}/>
+          <AddItem/>
+        </div>
+        <div className={styles.secondRow}>
+          <ListSort sort={sort} className={styles.listSort} {...props} />
```
lmorchard: Are ListSort or ListCounter used anywhere else separately? Might reduce complexity a little to combine this whole row into a single component? (Well, you know, other than the busy work to combine them, changes tests, etc 😅 )

src/list/manage/components/list-sort.js
```diff
@@ -0,0 +1,71 @@
...
+import { Localized } from "fluent-react";
+import PropTypes from "prop-types";
+import React from "react";
+
+import classNames from "../../../common";
+
+import styles from "./list-sort.css";
+
+export default class ListSort extends React.Component {
+  // TODO: do we need to do something with className?
```
lmorchard: Probably not, unless it's expected that someone using this will want to select a style variant. Could be YAGNI for now.

src/list/manage/components/list-sort.js
```diff
@@ -0,0 +1,71 @@
...
+
+  handleChange(evt) {
+    const value = evt.target.value;
+    this.setState({ value });
+    // TODO: onChange vs handleChange is confusing
+    this.props.onChange(value);
+  }
+
+  render() {
+    const { onChange, sort, className, disabled, ...props } = this.props;
```
lmorchard: Per earlier comment that onChange is confusing, I don't think you need onChange destructured here 😁 Same for className

src/list/manage/index.js
```diff
@@ -15,22 +15,29 @@ import reducer from "./reducers";
...
+import { saveSort, loadSort } from "./sort-middleware";
 
-const store = createStore(reducer, undefined, applyMiddleware(
-  thunk, telemetryLogger
-));
-store.dispatch(listItems());
-initializeMessagePorts(store);
-store.dispatch(getProfile());
+let store;
+(async () => {
```
lmorchard: Hmm, this *should* be near-immediate and not require any kind of visual loading indicator. But, wrapping the entire initialization in an async block waiting on the sort loading seems like it could be trouble.

Could you initialize the app, *then* start the async load and dispatch an action to update the state? (I guess that might cause a flash of content shifting around.) Also, is there a chance that an exception could get thrown here?

6a68:
Cleaned up all the unit test bustage due to changing the item summary and adding the sort state to the redux cache, and added tests for the counter and the sort utils. Just need to get the ListSort tests filled in, take a second look at the async init issue Les pointed out, and make sure I've hit all the other points from past reviews. I'll pluck the 'WIP' off this when it's good for a final pass.

src/list/manage/components/item-list-panel.js
6a68: Good questions! I think the ListCounter could abstract away the "foo entries found" code that's [currently inlined](https://github.com/mozilla-lockbox/lockbox-addon/blob/master/src/list/popup/components/item-list-panel.js#L72-L92) in the popup's ItemListPanel, though it'd be a bit of work to add a non-numeric default state to show the "recently used entries. click to search" text when there's no filter.

src/list/manage/index.js
6a68: More good questions! Re-sorting the list is going to be a highly visible change, so it seemed better to force it to wait a beat to hit localstorage.

There is indeed exception handling to deal with, but that happens in sort-middleware.js, in which case the `loadSort` return value is falsy, and we'll wind up going with the default sort.

6a68:
Whoops, not quite ready to land yet, looks like I've got a minor CSS bug:

Screen Shot 2019-03-14 at 9 26 10 AM

6a68:
Hmm. @linuxwolf do you know if we have any mocks for a ghost town / no entries state for the doorhanger? Not quite sure what that _should_ look like

linuxwolf:
> Hmm. @linuxwolf do you know if we have any mocks for a ghost town / no entries state for the doorhanger? Not quite sure what that _should_ look like

The `test/unit/list/manage/mock-redux-state.js` exports an `initialState` with no entries.  Should pull that up a couple levels, eventually.

6a68:
@linuxwolf heh, I meant UX mocks, not testing mocks ;-)

linuxwolf:
> @linuxwolf heh, I meant UX mocks, not testing mocks ;-)

🤦‍♂️ sorry!  The closest we have are "no matching entries"

* [invision](https://mozilla.invisionapp.com/share/ENOL84959AK#/screens/342865705)
* [Zeplin](https://zpl.io/agP3rj9)

6a68:
Fixed the last little CSS nit, should be ready to land

karma.conf.js
```diff
@@ -30,6 +30,9 @@ module.exports = (config) => {
         { type: "html", subdir: "html" },
       ],
     },
+    mochaReporter: {
```
linuxwolf: noice

linuxwolf:
In running this branch, I noticed things get out of shape if the viewport width is small enough:
image

Looking a little deeper, it appears the left-side list doeesn't have a minimum width.  Do we want to try and tackle that as part of this PR, or file a follow up?  I lean a little more toward a follow up, because I think other things might be affected (e.g., #99).

src/list/manage/components/item-list-panel.css
```diff
@@ -2,17 +2,51 @@
...
 }
 
 .item-filter {
   composes: item-filter from "../../../widgets/panel.css";
   background: #f3f3f5;
   height: 32px;
 }
+
+.list-sort {
+}
```
linuxwolf: This is present because of the imported CSS is expecting it, right?  If so, we probably want a comment so future us doesn't mistakenly remove it.

src/icons/arrow-dropdown.svg
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" ?><svg height="48" viewBox="0 0 48 48" width="48" xmlns="http://www.w3.org/2000/svg"><path d="M14 20l10 10 10-10z"/><path d="M0 0h48v48h-48z" fill="none"/></svg>
```
linuxwolf: The rest of our SVG icons are internally sized correctly, but it wasn't working out initially for this one.  I think we can change the `width` and `height` here, and maybe ditch setting the dimensions in CSS.

test/unit/background/message-ports-test.js
```diff
@@ -115,6 +127,15 @@ describe("background > message ports", () => {
       item,
     });
 
+    expect(result.item.timePasswordChanged < Date.now());
+    expect(result.item.timePasswordChanged > originalTime);
+
+    // Now, delete timePasswordChanged everywhere, so the items will be
```
linuxwolf: 😞  but understandable

src/list/manage/components/item-list-panel.css
6a68: I think we can just get rid of it

6a68:
@linuxwolf Thanks for the final pass! :beers:

Looks like `list/manage/components/app.css` does have a min-width of 150px on the `aside` class. Cranking it up to 250px fixes the collapsing list at narrow sizes, though the rest of the page still needs some min-width help:

Screen Shot 2019-03-14 at 4 00 28 PM

Fixed the other nits you pointed out. Merging.

PR #98, opened 2019-03-08
Build and Sign with CircleCI

### LINKED ISSUE #44: Investigate CI pipeline integrations with autograph service
Determine what it would take to pass through the autograph service and integrate with our build process.

Potentially involves switching from TravisCI to CircleCI.

## Acceptance Criteria
* Approach for automatically publishing a Mozilla-signed addon is documented
* Follow-up issues to track making the necessary changes are created

meandavejustice:
Fixes #44


.circleci/config.yml
```diff
@@ -0,0 +1,215 @@
...
+            chmod +x geckodriver
+            sudo mv geckodriver /bin
+            geckodriver --version
+      - run:
+          name: Install Firefox
+          command: |
+            sudo npm i -g get-firefox
+            get-firefox --branch nightly --platform linux --extract --target ~/
+            ~/project/firefox -v
+            firefox --version
```
linuxwolf: if I'm reading the above command correctly ...

1. `get-firefox` is extracts/instals to `~/`
2. attempts to run firefox to get its version from `~/project/firefox` (which doesn't exist)
3. attempts to run system-installed firefox to get its version

.circleci/config.yml
```diff
@@ -0,0 +1,217 @@
...
+      - run:
+          name: "Publish Release on GitHub"
+          command: |
+            VERSION=$(cat package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g' | tr -d '[[:space:]]')
+            echo ${VERSION}
+            echo ${GITHUB_TOKEN}
+            echo ${CIRCLE_PROJECT_USERNAME}
+            echo ${CIRCLE_PROJECT_REPONAME}
+            echo ${CIRCLE_SHA1}
+            ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -delete ${VERSION} ./project/signed-prod.xpi
```
linuxwolf: If I've followed everything correctly, the signed addon is located at `./project/addons/signed-prod.xpi` (rather than `./project/signed-prod.xpi`).

The error message from `ghr` is misleading to me, but I think it's happening because the path doesn't exist, therefore the caller forgot an argument, and therefore tells the caller to supply two arguments.

.circleci/config.yml
```diff
@@ -0,0 +1,218 @@
...
+              only:
+                - master
+      - docs-deploy:
+          requires:
+            - build-documentation
+          filters:
+            branches:
+              only:
+                - master
+      - package-addon-staging:
```
linuxwolf: I think we can get rid of this job

.circleci/config.yml
```diff
@@ -0,0 +1,218 @@
...
+          requires:
+            - build
+          filters:
+            branches:
+              ignore: /^\d+\.\d+\.\d+$/
+      - package-addon-production:
+          requires:
+            - build
```
linuxwolf: this job should depend on `test` instead of `build`

.circleci/config.yml
```diff
@@ -0,0 +1,218 @@
...
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
+      - run:
+          name: install-npm
+          command: npm ci
+      - save_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
+          paths:
+            - ./node_modules
+      - run:
```
linuxwolf: I don't think step is necessary; `geckodriver` (and selenium et al) are retrieved via `npm ci` and the `webextensions-geckodriver` dependency.

.circleci/config.yml
```diff
@@ -0,0 +1,218 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/lockbox-addon
```
linuxwolf: I don't know that this has any effect; elsewhere we seem to use `~/project`

.circleci/config.yml
```diff
@@ -0,0 +1,218 @@
...
+          key: dependency-cache-{{ checksum "package.json" }}
+          paths:
+            - ./node_modules
+      - run:
+          name: Package addon for production
+          command: |
+            npm run package
+            npm run sign:prod
+      - store_artifacts:
+          path: ~/project/addons/signed-prod.xpi
```
linuxwolf: maybe add a `destination` for a simplified `signed-addon.xpi` or `lockbox-signed.xpi`?

PR #104, opened 2019-03-15
Make CI Fail on Enforced Lint Errors

linuxwolf:
Found while working on an earlier patch; enforced linting is not part of the build, which can mean some classes of errors won't trigger a failure.

## Testing and Review Notes

Changing the code so that it would still function, but JavaScript or CSS linting would produce errors.

## To Do

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,6 +182,7 @@ workflows:
           requires:
             - build
             - test
+            - lint
           filters:
             branches:
               ignore: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,27 @@ jobs:
           root: .
           paths:
             - ./*
+  lint:
+    docker:
+      - image: circleci/node
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: update-npm
+          command: 'sudo npm install -g npm@6.7'
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
+      - run:
+          name: install-npm
+          command: npm ci
+      - save_cache:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -61,6 +61,9 @@ jobs:
           command: npm run integration
           environment:
             LOCKBOX_FIREFOX_HEADLESS: "1"
+      - run:
```
lmorchard: This *might* be worth running first before any tests, so things fail fast on lint issues

linuxwolf: That's a very fair point.  Personally, I've cycled between before, after, or parallel.  I think I pushed up after so that we see functional failures before stylistic failures; would be just as content with stylistic first.

@6a68 @meandavejustice do you have a preference?

meandavejustice: either before or parallel seem optimal to me

linuxwolf:
Latest version changes to a parallel lint + test, and adds lint as a requirement for `package-addon-production` .

.circleci/config.yml
lmorchard: I would say *not* parallel, because I'd think you'd *want* this to preempt other tests and fail early. But 🤷‍♂️

linuxwolf: Those are fair points.  What is currently in this branch is a parallel test.  The Travis-CI version ran lint checks in parallel with tests, for that's worth.

If we do go with this, we can mark the `lint` job as required, which would block merging (though an admin could override).

6a68: I'd probably prefer to run everything in parallel, so we can gather as much diagnostic info as possible from each test run. But that's not a strongly held opinion.

PR #111, opened 2019-03-19
Inform user of sync errors

### LINKED ISSUE #61: Inform user of Sync/FxA errors
Child of #38 

If the user is logged in to Sync, there are two edge cases we might bump into:

1) their email is unverified
2) their password was changed on another device, so the locally-cached password is stale, and they need to sign in again to sync

In both cases, the existing desktop code redirects the user to the sync preferences page (about:preferences#sync) to fix their situation.

## Acceptance criteria
* Should redirect to the same endpoint
* Should alert the user when the email is unverified
* Should alert the user when the Account password changed
* Should define own entrypoint in telemetry

Open questions:
- Product: Do we want to just redirect to that same endpoint? Assuming yes, more questions:
- Eng: We should be able to open the correct link via either `gSync.openPrefs` or `gSync.openSignInAgainPage()` ([usage example](https://searchfox.org/mozilla-central/rev/c21d6620d384dfb13ede6054015da05a6353b899/browser/base/content/browser-sync.js#279-283)) from an embedded API experiment. Need to make sure `gSync` is available in an API experiment js file (it should be).
- UX: What kind of UX do we want to offer in Lockbox for those cases? Some kind of info screen and a button to redirect to the preferences page? Show an alert badge over the icon?
- Telemetry question: The existing desktop hamburger menu item appends a query string when sending users to the sync preferences page: "about:preferences?entrypoint=menupanel#sync". Do we need to register a new entrypoint? (markh's thinking was that this query string is part of fxa server telemetry, ingested via amplitude, so, maybe it's dynamic and "entrypoint=lockbox" doesn't need to be explicitly registered anywhere)

### Visual Design

**InVision**
Full View: https://mozilla.invisionapp.com/share/ENOL84959AK#/343774936_Logins_-_Sync_Issue_-Light-
Full View - User Menu: https://mozilla.invisionapp.com/share/ENOL84959AK#/343989952_Logins_-_Sync_Issue_-_Menu_-Light-
Doorhanger: https://mozilla.invisionapp.com/share/ENOL84959AK#/343777425_Doorhanger_-_Sync_Issue_-Light-

**Zeplin**
Full View: https://zpl.io/2ZjOwLJ
Full View - User Menu: https://zpl.io/adrod9e
Doorhanger: https://zpl.io/VOKELQl

*Link would direct to sync preferences

Requires #92

meandavejustice:
Fixes #61 

## Testing and Review Notes

You may test by opening up accounts.firefox.com in a private window and changing your password.

## Screenshots
popup-error-sync

### DIFF
--- a/src/list/manage/containers/app-header.js
+++ b/src/list/manage/containers/app-header.js
@@ -218,10 +218,10 @@ export class AppHeader extends React.Component {
 export default connect(
   ({
     app: {
-      tabsReducer: {
+      tabs: {
         selectedTab,
       },
-      profileReducer: {
+      profileWrap: {
         profile,
         hasProfile,
       },
--- a/src/list/manage/containers/app-header.js
+++ b/src/list/manage/containers/app-header.js
@@ -218,9 +218,13 @@ export class AppHeader extends React.Component {
 export default connect(
   ({
     app: {
-      selectedTab,
-      profile,
-      hasProfile,
+      tabsReducer: {
+        selectedTab,
+      },
+      profileReducer: {
+        profile,
+        hasProfile,
+      },
     },
   }) => ({
     selectedTab,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1417,6 +1417,554 @@
             "strip-eof": "^1.0.0"
           }
         },
+        "fsevents": {
+          "version": "1.2.4",
+          "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.4.tgz",
+          "integrity": "sha512-z8H8/diyk76B7q5wg+Ud0+CqzcAF3mBBI/bA5ne5zrRUUIvNkJY//D3BqyH571KuAC4Nr7Rw7CjWX4r0y9DvNg==",
+          "dev": true,
+          "optional": true,
+          "requires": {
+            "nan": "^2.9.2",
+            "node-pre-gyp": "^0.10.0"
+          },
+          "dependencies": {
+            "abbrev": {
+              "version": "1.1.1",
+              "bundled": true,
+              "dev": true,
+              "optional": true
+            },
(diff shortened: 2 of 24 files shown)


meandavejustice:
Just noticed the weirdness of the top border of the notification in the popup, it's abt 1 pixel off in the center section. Gunna investigate that and update before removing the "WIP", should be good for review otherwise.

meandavejustice:
Inkedfixed-pop_LI

meandavejustice:
@linuxwolf I think I've addressed your comments in https://github.com/mozilla-lockbox/lockbox-addon/pull/111/commits/04885bf17ac77386072b8e20c11e3819cedb095f, I renamed `profileReducer` => `profileWrap` in order to avoid a confusing structure like: `app.profile.profile`

src/list/manage/containers/app-header.js
```diff
@@ -218,10 +218,10 @@ export class AppHeader extends React.Component {
 export default connect(
   ({
     app: {
-      tabsReducer: {
+      tabs: {
         selectedTab,
       },
-      profileReducer: {
+      profileWrap: {
```
linuxwolf: this makes sense

changecourse:
Happy to suggest that we shorten to just "Reconnect" if we think this line break would be experienced by a decent percentage of users.

linuxwolf:
> Happy to suggest that we shorten to just "Reconnect" if we think this line break would be experienced by a decent percentage of users.

while very anecdotal, the profile I used is a dev/testing one -- the only changes I deliberately made after creating it are Sync- and logins-related.  It might that a decent number of our Mac users will see it this way.

meandavejustice:
>he button wraps to another line in the doorhanger:
ah I didn't see that on my machine. I could switch the font-size to use rem instead of px

PR #112, opened 2019-03-20
Promotional Banners

### LINKED ISSUE #14: Promote Sync on mobile Lockbox apps
Child of #38

When I want to get my passwords across my many devices, I want to clearly know how to do that

## Acceptance criteria
* Should indicate in the password management interface that the user can download companion mobile apps
* Should not be possible (yet?) to dismiss banner
* Should open the call-to-action in a new tab/window
* "Learn More" link should be the only clickable area
* Should always display regardless of the Sync state of the user

## Implementation Details
* Should be prepared to be contextual

## Visual Design
**Zeplin**
https://zpl.io/boK9ggk
(Current approach is to link to lockbox.firefox.com page for both iOS and Android, letting the functionality built on that page handle text/email of app, as well as providing additional information to the user regarding the apps)

### LINKED ISSUE #79: Promote FxA/Sync sign in when user is not logged in
When I'm not signed in with a FxA to Sync (specifically Sync for logins), I need to first know how to do that.

## Acceptance criteria
* Sign in to FxA and Sync should be the primary call to action if someone is not already
* Should promote the sign in from the top navigation

To consider: do we want to change the promotion at the bottom of the screen when not signed in?

## Visuals
Logins - Home, unauthenticated (Light)

related #78
child of #38

linuxwolf:
Displays a promotional banner in the management interface when a user has an item selected:

* If the user is signed in --> "Lockbox on Mobile"
* If the user is **not** signed in --> "Sign up for FxA"

The implementation created a very basic banner widget we can extend to support closing, then specific promotional banner sub-components for device and fxa, respectively.

Fixes #14 
Fixes #79 

## Testing and Review Notes

To see the banner:
1. open Lockbox management interface
2. select an item

The banner should always be rendered.  As with other aspects of Lockbox, the implementation is responsive to state changes in Firefox; signing in (or disconnecting) will change the banner -- even while the page is still loaded.

* **not signed in**: Clicking "Sign In" takes the user to the Sync prefs page to start the signin process.
* **signed in**: Clicking "Learn More" takes the user to the Lockbox website, which will (very _very_ soon) promote the mobile apps.

## Screenshots or Videos

**not signed in**
image

**signed in**
image

## To Do

- [x] add unit tests
- [ ] check on the [accessibility](https://mozilla-lockbox.github.io/lockbox-addon/developer/test-plan-accessibility/) of any added UI


changecourse:
Small typo: "Createa a..." on first banner.

src/locales/en-US/common.ftl
```diff
@@ -2,8 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+product-line = Firefox
 product-title = Lockbox
 product-tagline = The simple way to store, retrieve and manage website login info
+product-fulltitle = { product-line } { product-title }
```
6a68: We should check with the content / branding folks about localizing the full title, either now or in a followup bug (and before we turn our localizers loose on this project). Product names are often [not translated at all](https://github.com/mozilla-services/screenshots/blob/master/locales/en-US/server.ftl#L69), and if that's the decision here, we may want to add a note above each of product-line, product-title, and product-fulltitle, asking localizers not to translate these, like:

```
# Note: please don't translate Firefox
product-line = Firefox
```

It's probably safe to assume these won't be translated, and add the notes now.

src/list/manage/components/promote-banner.css
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+.promotion {
+    min-width: max-content;
+    display: flex;
+    flex-flow: row nowrap;
+    align-items: center;
+    justify-content: baseline;
+}
+.promotion .content {
```
6a68: You might want to set a max-width on the `.content` div, so that the detail text wraps nicely in German and other languages that tend to have lengthy translations

src/locales/en-US/common.ftl
linuxwolf: ah, very good point.  We've done that in Android; haven't done that here yet because the Frech translation was an outside contribution.

src/list/manage/components/promote-banner.css
linuxwolf: I put a max-width on the overall banner, but max-width here would keep other layout in a better state.

linuxwolf:
With an IRL 👍 from @meandavejustice and a signoff from @6a68 , will be merging presently.

PR #113, opened 2019-03-21
🏋️‍♂️ telemetry

### LINKED ISSUE #29: Add telemetry events for relevant actions in the extension

6a68:
Fixes #29.

- couple broken tests left, couple tests to be written, and need to
finish implementing the code to measure password reveal events. but
definitely far enough along to get comments.

- opted to keep the telemetry calls in the redux middleware, and added
some actions that don't affect the redux store, but are only used by
the telemetry middleware. seemed cleaner than giving various react
components/containers knowledge of the telemetry library.

- refactored the telemetry methods to use an options object instead of
positional args. In our new metrics, value is optional and sometimes
null, and extra is optional and sometimes null; and it seemed that
getting the positional order right could be bug-prone.

- converted from snake_case to lowerCamelCase because the Telemetry C++
code doesn't seem to permit underscores in method or object names.
Also removed the dash in the "lockbox-addon" category name, which caused
similar breakage. "lockboxv1", in explicitly versioning the telemetry
schema, seems like a better choice to me, but I suppose we'll be able to
filter based on the addon version anyway, so perhaps this doesn't really
matter at all.

- added embedded telemetry API experiment to work around recordEvent
bug, but haven't yet written the tests. also need to test the fallback
correctly flips based on version number from within the recordEvent
webextension code.

- in process (but not done, and broken): adding an onReveal method to
trigger an action so we can count the password reveal clicks via telemetry

- removed the reconnecting_sync event, because we don't yet show any UI
in the FxA menu in the error case

- I set `record_on_release` to `true` for everything, because our users
are opting in by installing the addon (or clicking a button to enable
it). is this definitely okay with legal and privacy? it was a good
enough argument in the test pilot era, but things may have changed.

- added an event for clicks on the sort UI, with the selected option in
the extra field. will having one ping with an `extra` param with the key
`sort` instead of `itemid` mess up the data ingestion scripts? If so, I
can just split into 3 events with no value.

Fixes #???

_(**Required**: this reference (one or many) will be closed upon merge. Ideally it has the acceptance criteria and designs for features or fixes related to the work in this Pull Request.)_

Connected to #???

_(Optional: other issues or pull requests related to this, but merging should not close it)_

## Testing and Review Notes

_(**Required**: steps to take to confirm this works as expected or other guidance for code, UX, and any other reviewers)_

## Screenshots or Videos

_(Optional: to clearly demonstrate the feature or fix to help with testing and reviews)_

## To Do

- add “WIP” to the PR title if pushing up but not complete nor ready for review
- [ ] add testing notes and screenshots in PR description to help guide reviewers
- [ ] add unit tests
- [ ] request the "UX" team perform a design review (if/when applicable)
- [ ] check on the [accessibility](https://mozilla-lockbox.github.io/lockbox-addon/developer/test-plan-accessibility/) of any added UI


irrationalagent:
I'll take a closer look soon, but to answer one thing now:

>     * I set `record_on_release` to `true` for everything, because our users
>       are opting in by installing the addon (or clicking a button to enable
>       it). is this definitely okay with legal and privacy? it was a good
>       enough argument in the test pilot era, but things may have changed.

This should be fine but will be confirmed in data steward review. We are only collecting category 1 and 2 data and I believe that's the most important criteria

linuxwolf:
>     * "lockboxv1", in explicitly versioning the telemetry
>       schema, seems like a better choice to me, but I suppose we'll be able to
>       filter based on the addon version anyway, so perhaps this doesn't really
>       matter at all.

The explicit versioning here was to work around (former?) bugs in the C++ Telemetry implementation which would throw errors if the schema already existed; the number was (supposed to be) incremented each time a telemetry event/scaler was added or changed, and did not reflect the version of Lockbox.

6a68:
> > ```
> > * "lockboxv1", in explicitly versioning the telemetry
> >   schema, seems like a better choice to me, but I suppose we'll be able to
> >   filter based on the addon version anyway, so perhaps this doesn't really
> >   matter at all.
> > ```
> 
> The explicit versioning here was to work around (former?) bugs in the C++ Telemetry implementation which would throw errors if the schema already existed; the number was (supposed to be) incremented each time one or me telemetry events/scalars were added or changed, and did not reflect the version of Lockbox.

This was exactly what I was worried about with registering all the events at once. I'll restore the version number.

docs/metrics.md
```diff
@@ -124,23 +127,25 @@ These are the metrics we currently collect regarding the state of the user's log
...
 
-1. `startup` fires when the webextension is loaded. **objects**: webextension. Note that this event fires whenever the browser is started, so is not indicative of direct user interaction with Lockbox. **value** is null.
+1. `startup` fires when the webextension is loaded. **objects**: `webextension`. Note that this event fires whenever the browser is started, so is not indicative of direct user interaction with Lockbox. **value** is null.
 
-2. `click` fires when someone clicks on one of the UI elements listed **objects**: `toolbar`, `get_mobile`, `faq`, `account_settings`, `give_feedback`, `settings_menu`, `reconnect_sync`, `signin_sync` (or whatever the menu is that contains links to account settings, faq, etc). **value** is null.
+2. `click` fires when someone clicks on one of the UI elements listed **objects**: `toolbar`, `getMobile`, `faq`, `accountSettings`, `giveFeedback`, `settingsMenu`, `signinSync`.  **value** is null. `sortMenu` has **value** of `lastUsed`, `lastChanged`, or `name`.
```
irrationalagent: nice +1 for adding value for menu sort

docs/metrics.md
```diff
@@ -124,23 +127,25 @@ These are the metrics we currently collect regarding the state of the user's log
...
 
-3. `show` events fire when various UI elements are rendered shown to the user. **objects**: `item_list_manager`, `item_list_doorhanger`, `item_detail_doorhanger` `item_detail_manager`, `new_item`, `item_edit`, `delete_confirm`, `connect_another_device` (referring to the dialog displayed after a user clicks on the button to learn about the mobile apps) **value** should be a boolean for `item_list_*` events indicating whether any logins are in the login list (e.g. `False` if the list is empty).
+// TODO: do we just want true/false to indicate if the list is empty, or do we want to send the number of logins shown?
```
irrationalagent: I had just a bool because I'm trying to avoid sending precise values for now. There is a separate scalar that gets set by the browser that's a binned value for the # of passwords saved

docs/metrics.md
```diff
@@ -124,23 +127,25 @@ These are the metrics we currently collect regarding the state of the user's log
...
+3. `show` events fire when various UI elements are rendered shown to the user. **objects**: `itemListManager`, `itemListDoorhanger`, `itemDetailDoorhanger` `itemDetailManager`, `newItem`, `itemEdit`, `deleteConfirm`, `connectAnotherDevice` (referring to the dialog displayed after a user clicks on the button to learn about the mobile apps) **value** should be a boolean for `itemList*` events indicating whether any logins are in the login list (e.g. `False` if the list is empty).
 
-4. `item_add`, `item_update`, `item_delete` fire after a successful adding, editing or deleting of a credential. **objects**: manager, doorhanger (contingent on where the user initiated the action). item GUID should be in the extra field. **value** is null
+// TODO: note, users can't add, update, or delete from the doorhanger. should we keep or remove the object?
+4. `itemAdd`, `itemUpdate`, `itemDelete` fire after a successful adding, editing or deleting of a credential. **objects**: manager, doorhanger (contingent on where the user initiated the action). item GUID should be in the extra field. **value** is null
```
irrationalagent: let's keep the object == `manager` so that I can group by general events within that UX element

docs/metrics.md
irrationalagent: actually I see below you have a scalar that counts datastore.length or similar. Let's just keep that for now and see if anything comes up in data review.

src/background/telemetry.js
```diff
@@ -2,121 +2,189 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const TELEMETRY_CATEGORY = "lockboxv1";
+// TODO: dashes are not allowed in category names, so "lockbox-addon" is now "lockboxaddon"
```
irrationalagent: ✔️

src/background/telemetry.js
```diff
@@ -2,121 +2,189 @@
...
+        "itemListManager", // list logins in manage page, I guess each page load?
+        "itemListDoorhanger", // list logins in doorhanger, I guess each time the doorhanger's shown.
+        "itemDetailManager", // view item details in manage UI, per-login.
+        "itemDetailDoorhanger", // view details in doorhanger, per-login.
+        "itemEdit", // item edit view, per-login.
+        "deleteConfirm", // delete confirm modal, per-login
+
+        // not implemented:
+
+        "connectAnotherDevice", // TODO: not sure this makes sense as a "show" event, unless it's for the unimplemented mobile upsell (#97?)
```
irrationalagent: I think I wasn't sure what this actually referred to when I spec'd it. But if and when we have the mobile upsell, if its shown in all management views then we don't need a view event - we should definitely have some kind of `engage` method though.

src/background/telemetry.js
```diff
@@ -2,121 +2,189 @@
...
+      extra_keys: [],
+      record_on_release: true,
+    },
+    "openWebsite": {
+      methods: ["openWebsite"], // fires when a user clicks to open the website inside a login
+      objects: ["itemDetailManager", "itemDetailDoorhanger"],
+      extra_keys: [], // item GUID should be in the extra field. value is null.
+      record_on_release: true,
+    },
+    // TODO: this one is still in progress
```
irrationalagent: we don't need to land all the telemetry in one PR (unless that's what you want).

src/background/telemetry.js
```diff
@@ -2,121 +2,189 @@
...
+    "loginCount": {
+      kind: "count",
+      keyed: false,
+      record_on_release: true,
+    },
+
+    // not implemented yet
+
+    // Total number of logins copied or revealed during the current session
+    // TODO: increment a counter with each copy and each reveal
```
irrationalagent: thanks for planning to put this in, I know its extra work but its a valuable metric to have.

src/background/telemetry.js
```diff
@@ -2,121 +2,189 @@
...
-    return browser.telemetry.recordEvent(TELEMETRY_CATEGORY, method, object, null, extra || null);
-  },
-  scalarSet(name, value) {
-    return browser.telemetry.scalarSet(`${TELEMETRY_CATEGORY}.${name}`, value);
-  },
-};
+async function recordEvent({ method, object, extra = null, value = null }) {
+  let result;
+  try {
+    // Work around bug 1536877: use an embedded API experiment to submit
```
irrationalagent: FYI - this may be some time from now. Version uptake is kind of slow and we'd want 95+% of users to be at the threshold version.

src/background/telemetry.js
```diff
@@ -2,121 +2,189 @@
...
+
+function scalarSet({ name, value }) {
+  return browser.telemetry.scalarSet(`${TELEMETRY_CATEGORY}.${name}`, value);
+}
+
+export function initializeTelemetry() {
+  registerEvents();
+  registerScalars();
+
+  // Record the "startup" event at init time.
```
irrationalagent: does this fire on browser start even if the user doesn't open the manager? its OK if so I just want to be clear on it before I start using the data

src/list/manage/telemetry.js
```diff
@@ -3,49 +3,130 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
irrationalagent: I like the action => event mapping, and the fact you pulled them all in here rather than putting them all over the place. it will make it much easier for me to make changes on my own moving forward if need be without having to bug you all.

src/background/telemetry.js
6a68: Yeah, the addons are loaded at browser startup time

src/background/telemetry.js
6a68: :-P sad but true

6a68:
a heavy lift indeed. ready for review, one or two pings left to fixup. the embedded API is tested via the unit tests, rather than via a dedicated test page.

6a68:
I can bump up the coverage a bit before landing.

src/background/telemetry.js
```diff
@@ -4,119 +4,200 @@
...
-  },
-};
+// TODO: can value still default to null? I think so?
+async function recordEvent({ method, object, extra = null, value = "" }) {
+  let result;
+  // Note: 'extra' objects must have string values. Replace null with empty
+  // string, and convert numbers to strings.
+  if (extra) {
+    Object.keys(extra).forEach(key => {
+      if (extra[key] === null) {
```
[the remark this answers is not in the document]
6a68: yeah, I'll take a second look.

6a68:
Thanks for the review! I'll rebase, fix up the coverage and tests, and address the review comments, then merge.

6a68:
Hm. Almost done, just a couple TODOs left to resolve, but the `ConnectedProfileBanner` render tests are failing with `expected {} to have property 'propTypes'`. @linuxwolf any idea what I'm missing? Feels like something obvious I'm overlooking...

linuxwolf:
> but the `ConnectedProfileBanner` render tests are failing with `expected {} to have property 'propTypes'`

This reads like some breakdown in the properties (now?) expected on the connected React component.  I can take a look at the error to see if I have more insights ...

6a68:
Turns out the issue was that the test was doing `chai.use(chaiEnzyme)`, but was supposed to be doing `chai.use(chaiEnzyme())`. _le sigh_

Everything seems to be working. Filed one followup bug (#134) to rearrange the popup handler code so we can count toolbar button clicks.

I'll squash-merge once CI turns green :+1:

6a68:
This PR is so large and so old that I fear I've lost perspective...but the tests pass and the code seems to run ok. 🤞 merging

PR #123, opened 2019-04-02
Validation for password input

### LINKED ISSUE #18: Error Handling: "missing password" on add/edit save

meandavejustice:
- fixes #18
 

Looks pretty good after you submit and remove focus from the password input.

pw-errors-good

Looks not as good when password input is still focused after form is submitted:
pw-errors-bad

I don't know how much we would want to override default behaviors

### DIFF
--- a/src/list/components/item-fields.js
+++ b/src/list/components/item-fields.js
@@ -119,6 +119,14 @@ export class EditItemFields extends React.Component {
               maxLength: maxLength.toString()};
     };
 
+    const passwordErrorEl = (
+        <div className={`${styles.tooltip} ${styles.passwordErrorMsg}`}>
+          <div className={styles.arrowUp}></div>
+          <Localized id="item-fields-password-error-message">
+            <p>eRROr mESSAGe</p>
+          </Localized>
+        </div>);
+
     return (
       <div className={styles.itemFields}>
         <div className={styles.inputWrap}>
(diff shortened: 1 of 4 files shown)


linuxwolf:
> Looks not as good when password input is still focused after form is submitted:
> pw-errors-bad
> 
> I don't know how much we would want to override default behaviors

I fear there would be user confusion if the "saved logins" dropdown is displayed within the management UI.  If it is possible to override this from within the extension, let's do it.

src/list/components/item-fields.js
```diff
@@ -166,7 +174,8 @@ export class EditItemFields extends React.Component {
           <Localized id="item-fields-password">
             <LabelText>pASSWORd</LabelText>
           </Localized>
-          <PasswordInput className={styles.password}
+          <PasswordInput className={styles.password} required
+                         errorEl={passwordErrorEl}
```
linuxwolf: I'm curious why this approach was necessary, which is different from the approach taken for the hostname \<Input />.

meandavejustice: This is because the errorElement needs to be a sibling of the input, and passwordInput does some wrapping of the element

linuxwolf:
this works for me; it deviates from designs, but is a better value than overriding and rewriting the built-in validation logic.

PR #128, opened 2019-04-04
Duplicate entry notification

### LINKED ISSUE #110: No Indication if User happens to add a duplicate entry
If the following fields all identical, Firefox throws an error that is not handled:

* hostname (visible as "Website Address")
* submitFormURL (hidden)
* httpRealm (hidden)
* username (visible as "username")

Note this is most likely to happen when adding an entry, but could also occur when a user attempts to update an existing entry.

## Acceptance Criteria

* If a duplicate is found, a notification is displayed to the user that this entry already exists, with a link to navigate to the existing entry, however the user is left on the filled "New Entry" page in case they want to change either the username or the web address in order to clear the errors and create a new entry.

## Visual

**Invision:** https://mozilla.invisionapp.com/share/ENOL84959AK#/355565077_Logins_-_New_Entry_-_Filled-_Duplicate_Error_-Light- 
**Zeplin:** https://zpl.io/a7O65d8

meandavejustice:
Fixes #110 

In order to fully satisfy #110 we will need to create a method which will reverse lookup an item by its fields. This exceeds the time originally estimated for this issue. I've included the rest of the work up to that point. We should finish the work in a follow up commit (tracked as #129)

## Screenshots or Videos

Inkedunedited-dupe-login_LI

## To Do

- [X] double check the original issue to confirm it is fully satisfied
- [X] add testing notes and screenshots in PR description to help guide reviewers
- [ ] add unit tests
- [ ] request the "UX" team perform a design review (if/when applicable)


6a68:
> In order to fully satisfy #110 we will need to create a method which will reverse lookup an item by its fields. 

Two thoughts on how we could do this:

1. Firefoxy way: Add a simple passthru for `Services.logins.findLogins` to the `browser.experiments.logins` embedded API experiment, then call that to make sure we aren't creating a dupe. (Note that `Services.logins.findLogins` returns matching Logins, and the API experiment could convert those to plain JS objects before returning. I think we'd then point the user at the returned dupe with the most recent lastChanged or lastUsed date.)

2. Javascripty way: Since we already have all the user's logins (via `browser.experiments.logins.getAll()`), we could just filter the list by each field, but we'd have to get the wildcard behavior of `findLogins` just right--otherwise we won't detect the duplicate, but LoginManager will throw anyway. Seems potentially bug-prone.

The first option is probably better, since there are so many edge cases here. Happy to help with API questions.

src/list/manage/components/edit-item-details.js
```diff
@@ -59,9 +62,10 @@ export default class EditItemDetails extends React.Component {
   }
 
   render() {
-    const {fields: { title }, onSave, onCancel, onDelete, onReveal} = this.props;
+    const {fields: { title }, onSave, onCancel, onDelete, onReveal, error} = this.props;
     const {itemId, ...saveState} = this.state;
     const newItem = itemId === null;
+    const isDuplicate = error && !!~error.message.indexOf("This login already exists");
```
linuxwolf: it would be nice if we could better differentiate this error before it's thrown to the extension.

nit: prefer `...indexOf(...) === 0` over `!!~...`; the former, while more verbose, is also more explicit

linuxwolf:
Following up on the deep link: #129 is to track that as a follow-up.  @6a68 would you mind porting/copying your comment about potential solutions there?

src/list/manage/components/edit-item-details.js
linuxwolf: moved comment on differentiation to be a requirement for #129

linuxwolf:
Landing in master now.  Great work @meandavejustice !

PR #172, opened 2019-04-23
Set an item on management page load to avoid blank home page

### LINKED ISSUE #148: Initial install with sync enabled, blank manage page
When opening first opening the management interface, it displays the list of entries as expected, but a mostly blank page on the right side "details" pane.  Instead, it should display the first entry (if any).

### Steps to reproduce

1. open management page:
  - from "Lockbox" doorhanger
  - from "Logins & Passwords" in the main menu

* **Expected** The right side displays the first entry in the list from the left
* **Actual** The right side is blank (except for the promotional banner at the bottom)

## Considerations

* If the user does not have any saved logins at all, it should display the "getting started" instructional panels
* If the management interface is opened with a filter query that results in no entries, it should display the blank details pane

## Screenshot
initial-open-lockbox

meandavejustice:
- fixes #148

If there we have an item list, we select the first in the list based on the cached sort. This allows us to not have an empty view on load.

If there is a filter query in the url we still show the homepage.

default-view-woohoo

## To Do
- [x] add unit tests
- [x] make sure any new UI has telemetry events wired up and documented in docs/metrics.md, and verify that the events are recording properly (visit `about:telemetry#events-tab`, then choose 'dynamic' from the dropdown menu at top right, to view events fired by the addon

### DIFF
--- a/src/list/manage/index.js
+++ b/src/list/manage/index.js
@@ -44,26 +44,22 @@ let store;
 
   // if no filter check ensure that we have a default item selected
   const state = store.getState();
-  const all = state.cache.items;
-  let filteredItems = all;
+  const allItems = state.cache.items;
+  let selectedItems = allItems;
 
   if (state.list.filter && !state.list.filter.userEntered) {
-    console.log(`filter by ${state.list.filter.query}`);
     const filter = parseFilterString(state.list.filter.query);
-    filteredItems = all
+    selectedItems = allItems
       .filter((i) => filterItem(filter, i));
-    if (!filteredItems.length) {
-      filteredItems = all.slice();
+    if (!selectedItems.length) {
+      selectedItems = allItems.slice();
     }
-  } else {
-    console.log("no filter applied!");
   }
-  console.log(`sort by ${ state.cache.sort }`);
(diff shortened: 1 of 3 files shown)


src/list/manage/index.js
```diff
@@ -34,11 +36,22 @@ let store;
...
-  store.dispatch(listItems());
+  await store.dispatch(listItems());
   initializeMessagePorts(store);
   store.dispatch(getProfile());
   applyQueryFilter();
 
+  // if no filter check ensure that we have a default item selected
+  const params = new URLSearchParams(window.location.search);
+  const filter = params.get("filter");
+  if (!filter) {
```
linuxwolf: I think users will expect the first item from a query-based filter to be selected.

I think it's pretty doable, though it's probably going to feel like dupicate code from [all-items](https://github.com/mozilla-lockbox/lockbox-addon/blob/c15e67d23c46152b8d549f80d17b457ebba82526/src/list/manage/containers/all-items.js#L16-L22). 

Note that `state.filter.userEntered` will be false for a query-based filter.

linuxwolf:
@6a68 and/or @devinreams could you double-check my work?

PR #196, opened 2019-05-01
Update timeLastUsed on password field copy

### LINKED ISSUE #157: Copying password should update timeLastUsed
When the password field is copied, I think we should increment the `timeLastUsed` field in the LoginManager, but it doesn't look like we currently do so. _(edit: not `timePasswordChanged`, `timeLastUsed`)_

Probably could update this in `background/message-ports.js`, in response to the "copied_field" message.

6a68:
Fixes #157

## Testing and Review Notes

1. start lockbox via `npm run run`
2. create a new login manually in the manage UI
3. open the addon debugger via `about:debugging`, then get the original `timeLastUsed` via
```js
(await browser.experiments.logins.getAll())[0].timeLastUsed
```
4. copy the password field in the lockbox UI
5. check the `timeLastUsed` again; it should be updated

### DIFF
--- a/src/background/datastore.js
+++ b/src/background/datastore.js
@@ -188,6 +188,15 @@ class DataStore {
     }
     return item || null;
   }
+  async touch(id) {
+    const item = await this.get(id);
+    let touchedItem;
+    if (item) {
+      const touchedLogin = await browser.experiments.logins.touch(item.id);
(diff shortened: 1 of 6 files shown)


6a68:
Ah, think I got it - I wasn't mocking out the clipboard call

6a68:
OK, rebased and ready for final review.

Issue #206, opened 2019-05-03
Lockwise Rebranding


sandysage:
## Must Haves
* [ ] Update favicon/toolbar icon (branded)
* [x] Update logo (branded w/ name) #213
* [x] Update product name text #213
* [x] Update text for "entry"/"entries" --> "login"/"logins" #213
* [x] Window Title to full product name #213

## Nice to Haves
* [ ] (Management) Reflow Header Bar (#210)
* [ ] (Management) Reflow list Sidebar (#211)
* [ ] (Management) Reflow Details Panel (#212)

## References
* [Desktop Changes](https://docs.google.com/document/d/1fxgK7PDwnZodUW6XJO0zv7Ozm85DJgzDrmQtqe53XUs/edit#heading=h.woqebvspymga)
* Full Logo ([svg](https://drive.google.com/file/d/1QXyFFw6wXIjqBJnbxpViSvp-vt8F_ZmV/view))
* Icon (svg) (_TBD_)

6a68:
we're picking the highlighted items in here: https://docs.google.com/document/d/1fxgK7PDwnZodUW6XJO0zv7Ozm85DJgzDrmQtqe53XUs/edit#

PR #213, opened 2019-05-06
Update Branding Strings

linuxwolf:
Updates all the branding strings from Lockbox to Lockwise.  Also updates titles to be just the product, and (to match mobile designs) all instances of "entry" or "entries" to "login" or "logins" (respectively).

**NOTE**  the following are **unchanged**:
* telemetry identifiers (`lockboxv1` telemetry category, `lockbox-addon` prefs entrypoint)
* FAQ link targets (requires updates to content, which can be done in a way that preserves these)
* Survey/feedback links
* Icons (follow up PR once assets are available)

Connected to #206

## Testing and Review Notes

**Requires Manual Inspection/Verification**

## To Do


src/locales/en-US/list.ftl
```diff
@@ -121,13 +124,13 @@ intro-page-footer-copy =
   you can sync your information to this device by signing in to or
   creating a { -fxaccount-brand-name }. <go>Learn More</go>
 
-item-details-heading-new = Create New Entry
-item-details-heading-edit = Edit Entry Details
+item-details-heading-new = Create New login
+item-details-heading-edit = Edit login Details
```
6a68: Should this be capitalized, "Login"?

linuxwolf: d'oh, yeah it should

linuxwolf:
@6a68 
> However, searching for lockbox in the repo does turn up quite a few hits; the website links are probably the most important, but we can defer those if there's a separate ticket for updating the website. @linuxwolf what do you want to do with the leftovers?

Most of the website hits are a good callout, and we can update them now since https://lockwise.firefox.com is up and running (so to speak).  Will need to reconcile the targets that say "Lockbox", though.  That I'll make sure to create an issue for, and keep some amount of "backwards compatibility" (e.g., add an explicit `<a name="....">`).

The icons we're still waiting for, though.  I don't think it's worth creating follow up issues; we can check off the boxes in #206.

6a68:
Sounds good! :+1: from me, self-merge at your leisure

linuxwolf:
CC @irrationalagent 

A couple of open questions/confirmations:

1. The telemetry identifier is currently `lockboxv1`.  That should _**not**_ change, correct?
2. When opening Sync preferences, we use the telemetry identifier `lockbox-addon`.  Should that change to `lockwise-addon` or remain as-is?

irrationalagent:
> CC @irrationalagent
> 
> A couple of open questions/confirmations:
> 
>     1. The telemetry identifier is currently `lockboxv1`.  That should _**not**_ change, correct?

Correct, its ok to keep it the same. We will be filtering for webextensions only on the analysis side, so the data won't be confused with the bootstrapped addon.

>     2. When opening Sync preferences, we use the telemetry identifier `lockbox-addon`.  Should that change to `lockwise-addon` or remain as-is?

Are you talking about the value for `entrypoint`? if yes, let's stick with lockbox since that's what our historical data for the other apps is (and will remain)

linuxwolf:
> Are you talking about the value for `entrypoint`? if yes, let's stick with lockbox since that's what our historical data for the other apps is (and will remain)

Yes, it's the value for `entrypoint`.  Will keep it as-is then.

linuxwolf:
> Worth a version bump now or soon too?

Since we've followed the "increment version just before release" model here, I think it's too soon.

PR #218, opened 2019-05-08
Clean up l10n.toml

flodolo:
This way I'll get notified for changes, since I plan to start reviewing the current FTL files shortly.

### DIFF
--- a/l10n.toml
+++ b/l10n.toml
@@ -1,9 +1,5 @@
 basepath = "."
 
-locales = [
-  "fr-FR",
-]
-
 [[paths]]
   reference = "src/locales/en-US/*.ftl"
   l10n = "src/locales/{locale}/*.ftl"


flodolo:
I just realized that there's a CODEOWNERS in *docs*? That shouldn't work, as far as I know.

flodolo:
Side question, since I don't really understand React. Should "common" be removed from here? The popup file doesn't have it
https://github.com/mozilla-lockwise/lockwise-addon/blob/master/src/list/manage/index.js#L67

linuxwolf:
> I just realized that there's a CODEOWNERS in _docs_? That shouldn't work, as far as I know.

According to [GitHub](https://help.github.com/en/articles/about-code-owners#codeowners-file-location), `docs` is an acceptable location.

CODEOWNERS
```diff
@@ -0,0 +1 @@
+src/locales/en-US/*.ftl @flodolo
```
linuxwolf: Please remove this file; it is already present in `docs/`.

flodolo:
> According to [GitHub](https://help.github.com/en/articles/about-code-owners#codeowners-file-location), `docs` is an acceptable location. In practice, it works; we use this layout for just about all our projects and have CODEOWNERS honored.

The change landed [5 days ago](https://github.com/mozilla-lockwise/lockwise-addon/commit/919fd06227e799625112d9c7775af77ee8036e06), but this PR didn't trigger anything? Was it because I wasn't a collaborator?

linuxwolf:
> > According to [GitHub](https://help.github.com/en/articles/about-code-owners#codeowners-file-location), `docs` is an acceptable location. In practice, it works; we use this layout for just about all our projects and have CODEOWNERS honored.
> 
> The change landed [5 days ago](https://github.com/mozilla-lockwise/lockwise-addon/commit/919fd06227e799625112d9c7775af77ee8036e06), but this PR didn't trigger anything? Was it because I wasn't a collaborator?

I don't know exactly why that didn't.  It very well be because you were not yet a collaborator on the project.

Having `CODEOWNERS` in `docs` is definitely honored in our other repos, and has definitely been honored here.  If it doesn't for the next PR, we'll see what changes we should make.

PR #219, opened 2019-05-08
Fix l10n issues in FTL files

flodolo:
This addresses some of the comments that remained opened after the last changes.

- Fix coding style (e.g. spaces around placeables) and comments
- Added a Sync brand. That's what we [have in Firefox](https://transvision.mozfr.org/?recherche=-sync-brand-short-name&repo=gecko_strings&sourcelocale=en-US&locale=en-US&search_type=strings_entities), and it's used for the pane's name. It also clarifies when Sync is the product, or the verb.

### DIFF
--- a/src/locales/en-US/list.ftl
+++ b/src/locales/en-US/list.ftl
@@ -9,20 +9,25 @@
 ## - Declined to adapt to grammatical case.
 ## - Transliterated.
 ## - Translated.
+
 -firefox-brand-name = Firefox
 -product-short-name = Lockwise
 -product-full-name = Firefox Lockwise
+
 # “Account” can be localized, “Firefox” must be treated as a brand,
 # and kept in English.
 -fxaccount-brand-name = Firefox Account
 
-# All the following items are localizable.
+## All the following messages are localizable.
+
+-sync-brand-short-name = Sync
+
 header-logins-button = Logins
 header-app-title =
   .title = { -product-full-name }
 
 profile-menu-account = Account
-profile-menu-sign-in = Sign in to Sync
+profile-menu-sign-in = Sign in to { -sync-brand-short-name }
 profile-menu-connect = Connect a Device
 profile-menu-faq = FAQ
 profile-menu-feedback = Provide Feedback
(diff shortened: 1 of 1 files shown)


src/locales/en-US/list.ftl
```diff
@@ -191,13 +198,12 @@ simply-sign-in-other-device = Simply sign in to your { -fxaccount-brand-name } o
 download-mobile = Download the mobile app
 download-ios-android = { -product-full-name } is available on both iOS and Android. <learnmore>Click here</learnmore> to learn more and to send a link to your phone to download the app.
 before-access = Before you can access your logins on another device, you will need to connect a { -fxaccount-brand-name }.
-connect-a-firefox-account = Connect a { -product-full-name }
-connect-a-firefox-account-complete = Connect a { -product-full-name } (complete)
+connect-a-firefox-account = Connect a { -fxaccount-brand-name }
+# Displayed after the connection has completed
```
flodolo: I'm still not sure if this comment is correct.

src/locales/en-US/list.ftl
```diff
@@ -191,13 +198,12 @@ simply-sign-in-other-device = Simply sign in to your { -fxaccount-brand-name } o
 download-mobile = Download the mobile app
 download-ios-android = { -product-full-name } is available on both iOS and Android. <learnmore>Click here</learnmore> to learn more and to send a link to your phone to download the app.
 before-access = Before you can access your logins on another device, you will need to connect a { -fxaccount-brand-name }.
-connect-a-firefox-account = Connect a { -product-full-name }
-connect-a-firefox-account-complete = Connect a { -product-full-name } (complete)
+connect-a-firefox-account = Connect a { -fxaccount-brand-name }
```
flodolo: This was wrong in the file.

src/locales/en-US/list.ftl
6a68: It is; the string in the modal changes after you have signed in:

before:


Screen Shot 2019-04-29 at 10 01 02 AM


after:

Screen Shot 2019-04-29 at 10 03 28 AM

Issue #225, opened 2019-05-09
Style cleanups for 5/10 release


meandavejustice:
- [x] Sort Header – Arrow should be aligned to each label, rather than longest label
- [x] Search – Match active state to spec https://zpl.io/bWNeopM
- [x] Login Detail
  - [x] Login Detail – Match behavior of website address link to photon spec – https://design.firefox.com/photon/components/links.html 
  - [x] Login Detail – Update Edit Icon and icon colors `.item-summary__title___SnqUb` – set font weight to 500
- [x] Create New Login – Buttons misaligned and not sufficient padding, tooltip for website address not matching spec  https://zpl.io/andLDWA 
- [x] Create Login – match padding of Create Login (hr, header, and fields) to Login Detail
- [x] Edit Login – match padding of Create Login (header and fields) to Login Detail

meandavejustice:
cc @sandysage @changecourse

meandavejustice:
I'm working on the Sort Header section first

6a68:
@linuxwolf Hey, what's left to do for this bug?

linuxwolf:
> @linuxwolf Hey, what's left to do for this bug?

Updated the checkboxes to reflect things that are truly done.  Since we have a tiny bit left to get done (l10n), we can discuss whether to try and fiinish the rest of the checkboxes.

sandysage:
I'm moving the remaining checkboxes to another ticket so that they don't get lost. I think if the changes are approved then we can close this issue.

PR #232, opened 2019-05-10
sort header tweaks

meandavejustice:
refs #225 

- [x] Sort Header
  - [x] Sort Header – Height should be 33px, bg color should be #f9f9fa
  - [x] Sort Header – Typography styling should match https://zpl.io/bWNeopM 
  - [x] Sort Header – Arrow should be aligned to each label, rather than longest label


src/list/manage/components/list-sort.js
```diff
@@ -8,32 +8,47 @@ import React from "react";
...
+}
+
 export default class ListSort extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
       value: props.sort,
+      selectWidth: "65px",
     };
+    this.selectEl = React.createRef();
```
linuxwolf: This change is breaking the unit tests.  Not sure what kind of changes are needed in the tests to make it pass, since this is maintaining the right functionality.

PR #258, opened 2019-05-21
Add notes on localization preparation steps

devinreams:
Based on the fact Pontoon is hooked up to the `pontoon` branch now, we want to make sure we incorporate these steps to make sure translations get incorporated into the product.


devinreams:
For posterity, this is now based on a "real life" first run at #260

PR #259, opened 2019-05-21
Fix wrong term in sync-requires-account

flodolo:
I missed it while fixing the other strings


flodolo:
This is done on purpose against the branch used by Pontoon. It will need to be synced back to master once we have translations.

PR #260, opened 2019-05-21
Update strings from Pontoon: it

devinreams:
This not only includes a commit with the IT translation, but also includes the typo fix from #259 (from pontoon branch into master branch).

Documented here #258

### DIFF
--- a/src/locales/en-US/list.ftl
+++ b/src/locales/en-US/list.ftl
@@ -200,7 +200,7 @@ before-access = Before you can access your logins on another device, you will ne
 connect-a-firefox-account = Connect a { -fxaccount-brand-name }
 # Displayed next to connect-a-firefox-account string after the user has logged in.
 connection-complete = (complete)
-sync-requires-account = To sync your logins to another device, you will need to <signin>sign in or create a { -fxlockwise-brand-name }</signin>.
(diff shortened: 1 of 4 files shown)


linuxwolf:
I am now realizing that for reasons I can no longer remember, supported locales need to added to `src/locales/locales.json` in order to be usable in [production builds](https://github.com/mozilla-lockwise/lockwise-addon/blob/master/src/locales/README.md).

That, or change the assembly script to dynamically build them as they are for development builds.

devinreams:
>supported locales need to added to src/locales/locales.json

@linuxwolf Done. Thank you, I missed that entirely but equally love that it was in the README. Nice work 👏

devinreams:
@flodolo @linuxwolf I just tried changing my browser language to "it" and I didn't pick this up doing `npm run run`. I also changed my entire system to "it" and did not show the strings there either.

@linuxwolf Have we seen a translated version working/appear as expected recently? I do remember seeing French many months ago.. either way, I think we should still merge this in as-is but add locale testing to the QA plan. Can you please help see that through?

flodolo:
I tested one string the other day and it was working fine for me (installed from about:debugging as temporary)

devinreams:
Thanks for confirming @flodolo. @linuxwolf also confirmed it worked as expected for him. I must've made a mistake in how I loaded the extension.

PR #269, opened 2019-05-22
Disable failing integration/functional tests

linuxwolf:
The functional tests are constantly failing in odd ways.  Given the lifetime of this project and that manual QA has done more to find fuzzy issues than our automated tests, disabling the set.

No, this is not ideal – far from it.  However, these haven't been increasing quality, so their value is increasingly questionable.

### DIFF
--- a/test/integration/functional-test.js
+++ b/test/integration/functional-test.js
@@ -61,7 +61,7 @@ const mockLogin = {
 
 const WAIT_DELAY = 1000;
 
-describe("add-on UI", () => {
+describe.skip("add-on UI", () => {
   let webext, driver, helper, webdriver, By, until, selectors;
 
   const waitFor = by => driver.wait(until.elementLocated(by), WAIT_DELAY);


linuxwolf:
It looks like CircleCI is paused for some reason, but it's not obvious to me why.  If ok, may admin merge.  It is passing always for me locally, for what it's worth ...

**cc @devinreams**

devinreams:
>It looks like CircleCI is paused for some reason, but it's not obvious to me why

Yeah likely related to an incident with delayed notifications from GitHub to Circle CI this morning

>may admin merge. It is passing always for me locally, for what it's worth ...

Fine by me. These will run later and we'll know for certain after-the-fact. 👍

linuxwolf:
> > It looks like CircleCI is paused for some reason, but it's not obvious to me why
> 
> Yeah likely related to an incident with delayed notifications from GitHub to Circle CI this morning

And now I'm noticing that even "WIP" is waiting response 🤔

PR #276, opened 2019-05-23
Update strings from Pontoon, don't expose Italian yet

devinreams:
Pulling in the latest from the `pontoon` branch.

Plus, I will add a commit to remove "it" from the locales.json so it's not exposed in production builds yet until we discuss with UX and work out a plan next week.


src/locales/locales.json
```diff
@@ -1 +1,5 @@
+<<<<<<< HEAD
```
linuxwolf: looks like a conflict was merged

devinreams: Thanks for spotting, didn't "Command-S" enough in my open windows 😔

PR #283, opened 2019-05-28
Wrap URL uses to catch failures

### LINKED ISSUE #279: Hostnames without a Scheme cause Doorhanger to go blank
If a login entry has a `hostname` value that is not a URL (e.g., `www.example.com`, missing the scheme `https://`), the doorhanger is rendered completely blank.

### Original ###
I have a few recently used logins in the doorhanger. If I click one of them, the doorhanger just becomes completely white.

linuxwolf:
Fixes #279

Wraps calls to `new URL` in a try-catch (and using the original value as-is) in most places in order to prevent critical failures; in this case, preventing the popup (or management UI) from rendering/functioning at all.

## Testing and Review Notes

This one requires "bad data" to be present in the underlying stored logins.  Once present, attempt to view details for the relevant login.


linuxwolf:
> LGTM, add a unit test for the failure case and you're good to land

For some reason, `ItemFields` doesn't have any unit tests.  Added a test to the place that definitely makes the call into that section.

Issue #290, opened 2019-05-30
Lockwise can wrongly be installed on Firefox versions older than 67


SoftVision-CosminMuntean:
[Notes]:
- This seems to be a regression since we have verified that the add-on is not compatible with Firefox 66 and older in issue [#161](https://github.com/mozilla-lockwise/lockwise-addon/issues/161#issuecomment-491788570).
- If the Lockwise add-on was never installed on Firefox versions older than 67, the browser crashes after the first install attempt of the add-on. After the browser is restarted the Lockwise add-on works as expected and if it is removed and reinstalled the browser no longer crashes.
- Also since the add-on can be installed on Firefox versions older than 66 the users will encounter the old issue #161.

[Affected Versions]:
- Lockwise 2.2.2-alpha
- Firefox 66 and older

[Affected Platforms]
- All Windows
- All Mac
- All Linux

[Perquisites]
- Have the latest Firefox 66 version installed.
- Have a new clean profile.

[Steps to reproduce]:
1. Open the Firefox 66 version with the profile from prerequisites.
2. Navigate to "https://lockwise.firefox.com/" website.
3. Click the "Install for Firefox" button.
4. Observe the behavrior. 

[Expected results]:
- A doorhnager which informs the users that the add-on is not compatible with Firefox 66 is displayed.

[Actual results]:
- The add-on is installing then the browser crashes.

[Additional Notes]:
- Attached a screen recording of the issue:
![Lockwise on 66](https://user-images.githubusercontent.com/19226534/58622944-82771d00-82d5-11e9-989d-8415c1f2c86f.gif)

devinreams:
Odd. I do see the `[strict_min_version](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings)` still set in the manifest template and after downloading and unzipping the xpi...

6a68:
This seems extremely weird to me.

One possible workaround: we can sniff the user agent on the lockwise website and either disable or show an "upgrade your firefox" button if the version is less than 67.

6a68:
@SoftVision-CosminMuntean I'm able to reproduce on 66.0.5. Could you try a few older versions of firefox as well, and see if this is a regression or old, known behavior?

* after the initial install-crash on 66.0.5, I can restart, then go to about:addons and disable/enable or remove/"undo" without causing another crash.
[...]

PR #295, opened 2019-05-31
Disable Telemetry unless Forced On

linuxwolf:
Connected to #290

As uncovered as part of #290 ([BMO 15555734](https://bugzilla.mozilla.org/show_bug.cgi?id=1555734)), there are some circumstances where registering and recording telemetry events/scalars can cause the browser to crash.

As a mitigation until the root cause is resolved, this PR disables telemetry.

## Testing and Review Notes

1. Install the addon
2. Perform various actions
3. Open `about:telemetry` and observe dynamic events

There should be no events or scalars with the category `lockboxv1`.

## To Do

- [x] add unit tests

### DIFF
--- a/src/background/telemetry.js
+++ b/src/background/telemetry.js
@@ -14,6 +14,9 @@ const TELEMETRY_CATEGORY = "lockboxv1";
 //  which seems like a code smell, while the action / middleware approach solves
 //  this nicely.
 
+let enabled = false;
+function getTelemetryEnabled() { return enabled; }
+
 function registerEvents() {
   const events = {
     "startup": {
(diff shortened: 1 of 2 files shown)


src/background/telemetry.js
```diff
@@ -14,6 +14,9 @@ const TELEMETRY_CATEGORY = "lockboxv1";
 //  which seems like a code smell, while the action / middleware approach solves
 //  this nicely.
 
+let enabled = false;
+function getTelemetryEnabled() { return enabled; }
+
 function registerEvents() {
```
6a68: nit: since there's an enabled check at the start of all the other telemetry functions, you could add one at the top of `registerEvents` as well, just for uniformity

linuxwolf: I put the enabled check on all the exported functions, figuring that would be sufficient.

linuxwolf:
keeping open until a @mozilla-lockwise/product decision is made for go/no-go

PR #315, opened 2019-06-11
Account for LoginInfo Schema Changes

### LINKED ISSUE #309: Support LoginInfo IDL changes
Bug [1148205](https://bugzilla.mozilla.org/show_bug.cgi?id=1148205) changes the LoginInfo IDL to replace `hostname` with `origin`, and to replace `formSubmitURL` with `formActionOrigin`.

For compatibility with all versions of Firefox in the addon, we should allow for these keys in the embedded `logins` API experiment.

linuxwolf:
Fixes #309

This makes the necessary changes to the experimental API and datastore to support the schema changes being introduced  in Firefox 69.

## Testing and Review Notes

Everything should continue to work as it did, including unit- and integration-tests.

## To Do


src/experiments/logins/schema.json
```diff
@@ -17,12 +17,23 @@
           "optional": false,
```
6a68: This looks great! I think we need to make the hostname optional here in the schema, otherwise LGTM

linuxwolf: that's fair; it is enforced lower down anyway, which is for the best.

PR #323, opened 2019-06-14
Release 2.2.4-alpha

linuxwolf:


src/locales/locales.json
```diff
@@ -1 +1 @@
-["en-US", "cak", "cy", "de", "en-CA", "fr", "fy-NL", "hu", "it", "nl", "nn-NO", "pt-BR", "ro", "sv-SE", "vi", "zh-CN", "zh-TW"]
+["en-US", "cak", "cy", "de", "en-CA", "fr", "fy-NL", "hu", "id", "it", "kab", "nl", "nn-NO", "pt-BR", "ro", "sv-SE", "vi", "zh-CN", "zh-TW"]
```
devinreams: Nice! And since we're slowing down on releases, is now the time to consider adding the 90%+ translations too? Like "cs" and "el" and "sl"?

linuxwolf: I thought about it, but I don't know if there's any assurance they would get to 100%, and it can be very jarring to use an app that's _mostly_ translated

6a68:
Might be good to get a final check from @flodolo, too

src/locales/locales.json
flodolo: Define "slowing down"? I would expect them to reach 100% at this point, since we're not adding new strings. But it would be nice to make sure we ship an update in a couple of weeks.

docs/release-notes.md
```diff
@@ -1,5 +1,25 @@
 # Firefox Lockwise for Desktop Release Notes
 
+## 2.2.4-alpha
+
+_Date: 2019-06-14_
+
+### What's New
+
+* Localization support for Indonesian (id) is added.
```
flodolo: Since you're adding two languages, I'd couple of them out (Kabyle)

flodolo:
Also, Czech is at 100%. There are 4 warnings, and they're OK.

linuxwolf:
Thanks for chiming in @flodolo .  I'll update the release notes an supported locales accordingly.

"Slowing down" here means further releases of this add-on will only be for critical issues, though I whink we can have another for localization updates at the start of July.

flodolo:
> though I whink we can have another for localization updates at the start of July.

That would be definitely good. I will also stop new locales from opting in at this point.

Issue #188, opened 2019-04-30
Uninstall on in-tree supported platforms


linuxwolf:
Once this experience is available in Firefox, the extension should uninstall itself.

## Open Questions
* [ ] Fully automate, or inform + trigger?
* [ ] Sunsetting content?

PR #334, opened 2019-07-29
Fix Firefox max version to 69

linuxwolf:
This change, combined with an update to the updates manifest, defers to the built-in functionality in Firefox 70.

See [BMO 1560608](https://bugzilla.mozilla.org/show_bug.cgi?id=1560608)

Connected to #188
Connected to mozilla-lockwise/mozilla-lockwise.github.io#138

## Testing and Review Notes

This add-on should no longer install on Firefox versions 70 or higher.

## To Do

- [ ] add testing notes and screenshots in PR description to help guide reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "name": "lockwise",
   "id": "lockbox@mozilla.com",
   "version": "2.2.4-alpha",
-  "description": "Take your passwords everywhere. Note this functionality is built into Firefox 70 and above.",
+  "description": "Take your passwords everywhere. Note this functionality is built into Firefox 70 and above, so this add-on is capped to versions 69 and lower.",
   "engines": {
     "firefox": ">= 67",
     "node": ">=10.15.0"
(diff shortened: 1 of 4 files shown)


src/manifest.json.tpl
```diff
@@ -16,6 +16,7 @@
     "gecko": {
       "id": "{{id}}",
       "strict_min_version": "67.0",
+      "strict_max_version": "69.0",
```
mnoorenberghe: I believe this should be "69.*" to handle the dot-releases

package.json
```diff
@@ -3,7 +3,7 @@
   "name": "lockwise",
   "id": "lockbox@mozilla.com",
   "version": "2.2.4-alpha",
```
6a68: I'm not sure if manifest changes will be applied for existing users unless you bump the version number.

linuxwolf: it won't.  I wanted to get this mechanics done first, then do an actual release (including the latest from pontoon).

msujaws:
Any update here?

linuxwolf:
At least from local testing, this is doing what's asked, so merging.  Will then stage a releas with the latest updates from Pontoon.

